PR #47, opened 2015-01-07
Improve configure generation

hurlebouc:
Hello!

I think their is a failure in configure.ac.

 Here is the commit message:

*change the way the configure check presence of library FUSE : previous
one search twice (one for adding the path of the library to LDFLAGS and
the other for eventually checking that the library is actually added)
for the same symbol (fuse_new) that resuls in the configure script
caching the first result of the test (which is 'false' because the
library where not still added) and reuseing it in the second test. A
solution is to use two different symbols for theses two tests).


vgough:
Thanks, merged.

PR #51, opened 2015-01-11
Fix reverse tests on OSX, improve install documentation

rfjakob:


vgough:
Thanks, sorry for the delay.

Issue #65, opened 2015-03-14
Configure error on Centos 6.6


teodorg:
checking for sys/xattr.h... yes
checking whether xattr interface takes additional options... no
./configure: line 20551: syntax error near unexpected token `newline'
./configure: line 20551: `  yes:no:'

And these lines are:

    # So?  What about this header?
    case $ac_header_compiler:$ac_header_preproc:$ac_cxx_preproc_warn_flag in
      yes:no:
    fi

rfjakob:
Which encfs version?

teodorg:
The master branch.

rfjakob:
Looks like something went wrong in the autoconf stage. Please try this configure script: https://github.com/rfjakob/encfs/releases/download/v1.8/configure
For 1.8.1, I will bundle the configure script in the .tar.gz.

compevo:
We have the same problem on Centos 6.6 even with the new configure script above.
./configure does succeed but then it seems like "make" rewrites the same bad ./configure and then build fails.

checking whether xattr interface takes additional options... no
./configure: line 20551: syntax error near unexpected token `newline'
./configure: line 20551: `  yes:no:'
make: *** [config.status] Error 2

dougalb:
I am seeing the same issue. I am guessing it is something to do with autoconf/automake versions on RHEL/CentOS6. I have been able to build with no issues on Amazon Linux 2015.03.

vgough:
The autoconf scripts have been replaced with cmake on master.  You will probably have to download a recent cmake binary (http://www.cmake.org/download/) in order to create the make files, since the default from RHEL 6.6 is probably too old.

If you uncover issues with the cmake build rules, please open a new ticket with details.

Thanks!

PR #69, opened 2015-03-22
Changes for 1.8.1

rfjakob:
One thing that may be debateable is that I have disabled uniqueIV for reverse mode by default (added in v1.8). It can still be enabled through expert mode. I have explained the details in commit 21b3811f86d0803c30bd3baee34e1769d06597b1 , but it boils down to that it may be an unexpected change for users, and cause regressions for them, just like disabling the cache did.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -34,6 +34,8 @@ Makefile.in
 /po/remove-potcdate.sed
 /po/Makefile.in.in
 /po/Rules-quot
+/po/*.gmo
+/po/stamp-po
 
 /m4/*.m4
 !/m4/ax_boost_base.m4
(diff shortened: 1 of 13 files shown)


vgough:
Nice, thanks Jakob!

samrocketman:
This is great!  I'm glad to see the project continue.

rfjakob:
:)
Valient, I'd like to release 1.8.1, this time including the configure script in the tarball. You ok with that?

vgough:
Sure, although I don't think it is unreasonable to expect those building from source to be able to generate the configuration file, but I realize that autoconf can be messy.

Please document somewhere how you created the tarball, so it can be done consistently in future releases.

rfjakob:
Yes, otoh, this bug report is where the idea came from:
https://github.com/vgough/encfs/issues/65
I thought including the configure script in release tarballs will avoid a
lot of headache for our users. I will script the process.
If you want to build master, you'd still need autoconf.

PR #74, opened 2015-03-29
Release 1.8.2

rfjakob:
* Fix `make dist` so we can create complete release tarballs again
* Add `--no-location` option to xgettext to make po diffs a lot smaller in the future


vgough:
Ahh, was synced to the wrong point.  I'll submit a small change to move the whitespace out.  The po files will need to be regenerated.

PR #84, opened 2015-06-16
Replace autoconfig with CMake

vgough:
This replaces the autoconfig build system with CMake.

This is easier to work with, and produces better warnings when it fails.
This has been tested on OSX and Ubuntu (via a drone build).  

Note that there isn't a 'test' target yet, as that hasn't been ported.

@rfjakob : any concerns?


rfjakob:
Very nice. Tested on Fedora 22 as well. Bye bye m4!

vgough:
Great, thanks for checking.  I'm not sure why the perl scripts aren't working, so will leave them commented out for now.

rfjakob:
Ok. I will take care of the Perl scripts.

PR #102, opened 2015-08-04
Build fixes to compile on Linux Mint 17.2

cognitive-i:
Unfortunately encfs 1.7.4 shipping with Linux Mint 17.2 appears broken (Unable to find nameio interface nameio/block, version 4:0:0).

This pull request puts in two minor tweaks to the CMakeLists.txt to build with the versions of CMake (2.8.12.2) and g++ (4.8.4) that Mint ships with.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.1)
+cmake_minimum_required(VERSION 2.8)
 project(EncFS C CXX)
 
 set (ENCFS_MAJOR 1)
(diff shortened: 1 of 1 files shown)


pepa65:
Excellent! That should work on Ubuntu LTS as well I would think.

vgough:
This isn't a complete replacement for CMAKE 3.1's C++11 support.  Can you check if #105 works for you?  I have checked it on OSX and changed the Circle CI build rules to run it against default cmake (2.8 I believe).

vgough:
Dropping this.

PR #106, opened 2015-08-20
Use Git version in build

vgough:
Automatically creates a version based on the currently checked out code.


rfjakob:
So this is how the version string looks like with this patch:

    v1.8.1+63.da3442b

Somehow strange non-standard format and no dirty marker, but better than the status quo I guess.
This is what `git describe --tags --dirty` does:

    v1.8.1-53-gda3442b-dirty

rfjakob:
I don't really understand how adding a new version is supposed to work, though.

1) Tag a new version
2) Build -> This updates the VERSION file, hence changes the source code
3) ???

vgough:
@rfjakob yeah, it seems messy, so not sure it's worth the trouble. Do you know of a cleaner way to embed build info?

rfjakob:
How about including the .git directory in the release tarballs? Or a
shallow clone if it is too big.
Then we could completely drop the version file and just rely on git
describe.

rfjakob:
Disregard that, too complicated.
Let's just not check in the VERSION file and only include it in release
tarballs.

PR #107, opened 2015-08-20
add xattr translation

### LINKED ISSUE #70: Problem opening files on OS X
This is a rare issue, rare enough to be difficult to reproduce, but not rare enough to not be annoying. We use version 1.8 installed from homebrew. 

How to reproduce (when this happens):
1. Double-click on the file (in this case PDF, but same happens with Word for example)
2. Dialog window saying "File_1" is damaged and can't be opened. You should eject the disk. It also says "Preview created this file on {date}.", which is correct. There are two options: Cancel and Eject Disk

If I copy this file to /tmp, I can open it, they are even same:

```
$ md5 encfs/File_1.pdf
MD5 (encfs/File_1.pdf) = 0bc06a6fae03f2ca9c4c55cd7cd81ad4
$ md5 /tmp/File_1.pdf
MD5 (/tmp/File_1.pdf) = 0bc06a6fae03f2ca9c4c55cd7cd81ad4
```

The underlying folder is synchronized using Synology CloudStation (private Dropbox alternative), so this may be related. It does not happen with unencrypted folder, so it's either an encfs issue or encfs+Synology CloudStation interaction issue. Internally, it's like dropbox, synchronizes the underlying files when they change (so it's not using FUSE or mount any other filesystem, it's an userspace application). Also noted: This file has been synchronized months ago and has not been changed, so it's probably not a locking issue and for sure not a synchronization conflict issue.

vgough:
Translate extended attributes with special meaning on OSX.

This is based on code from OSXFuse's loopback example.  Intended to fix #70

### DIFF
--- a/encfs/encfs.cpp
+++ b/encfs/encfs.cpp
@@ -17,20 +17,25 @@
 
 #include "encfs.h"
 
+#include <cerrno>
+#include <cstddef>
+#include <cstdio>
+#include <cstdlib>
+#include <cstring>
 #include <fcntl.h>
 #include <inttypes.h>
+#include <memory>
 #include <stdint.h>
+#include <string>
 #include <sys/stat.h>
 #include <sys/statvfs.h>
 #include <sys/time.h>
 #include <time.h>
 #include <unistd.h>
+#include <utility>
 #include <utime.h>
-#include <cerrno>
-#include <cstddef>
-#include <cstdio>
-#include <cstring>
-#include <memory>
+#include <vector>
+
 #ifdef linux
 #include <sys/fsuid.h>
 #endif
@@ -43,9 +48,6 @@
 
 #include <rlog/Error.h>
 #include <rlog/rlog.h>
-#include <functional>
-#include <string>
-#include <vector>
 
 #include "Context.h"
 #include "DirNode.h"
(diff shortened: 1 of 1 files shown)


vgough:
TODO: com.apple xattrs need to be thrown out when decoding, so that they aren't passed through.  This will effectively strip com.apple xattrs from existing filesystems though.

encfs/encfs.cpp
```diff
@@ -60,18 +62,21 @@ namespace rlog {
...
 using namespace std::placeholders;
 using namespace rlog;
 using rel::Lock;
 
-#define GET_FN(ctx, finfo) ctx->getNode((void *)(uintptr_t) finfo->fh)
+#define GET_FN(ctx, finfo) ctx->getNode((void *)(uintptr_t)finfo->fh)
+
+// XAttr prefix replacement.
+// Replaces "com.apple." with "encfs_apl." when storing xattrs.
+// TODO: future versions of encfs should encrypt xattrs, making this
```
vgough: skip this and go straight to xattr encryption?

benrubson: Yep.
Did we ever encounter (other) occurrences of the addressed issue ?

benrubson: Yes, in addition sounds like #70 is fixed since FUSE for OS X (2.8.2).
So I think we can close this PR yes.

PR #125, opened 2015-11-12
Perform checks to ensure that we're not descending into ourself.

wdoekes:
It's possible to mount this filesystem in a descendant of the real
(source) filesystem. For instance, one could do this:

    encfs --reverse / /home/encrypted/rootfs

At that point, all files in `/` (like `/root/.bashrc`) are also in
`/home/encrypted/rootfs` (like `/home/encrypted/rootfs/root/.bashrc`).

This can be useful when you want to export an encrypted copy of the
filesystem: the remote backup machine can fetch any file, but all files
will be encrypted.

However, the mountpoint itself is also there:

    /home/encrypted/rootfs/home/encrypted/rootfs

This would cause a `find /` of the filesystem to take infinite time. And
what's worse; trying to read files from there would cause the filesystem
to lock up:

    cat /home/encrypted/rootfs/home/encrypted/rootfs/root/.bashrc
    (infinite hang)

This patch adds an extra check so the filesystem refuses to descend into
itself.

### DIFF
--- a/encfs/DirNode.cpp
+++ b/encfs/DirNode.cpp
@@ -254,12 +254,8 @@ DirNode::DirNode(EncFS_Context *_ctx, const string &sourceDir,
   Lock _lock(mutex);
 
   ctx = _ctx;
-  rootDir = sourceDir;
+  rootDir = sourceDir; // .. and fsConfig->opts->mountPoint have trailing slash
   fsConfig = _config;
-
-  // make sure rootDir ends in '/', so that we can form a path by appending
-  // the rest..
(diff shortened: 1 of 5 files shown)


rfjakob:
Looks like a good idea! I have added a few comments to your patch ( https://github.com/ossobv/encfs/commit/6d9f16f1da68caea4e58e55d333464d633d0b230 ) .

wdoekes:
Thanks for the review! Updated with the requested changes.

rfjakob:
Looks good now! @vgough OK with merging this?

vgough:
lgtm, thanks.

PR #127, opened 2015-11-12
Repair tests

rfjakob:
The test suite has been broken since the switch to cmake. Repair it. All tests pass at the moment.


vgough:
Thanks!

Issue #112, opened 2015-09-12
[bisected] EncFS does not respect POSIX permissions on files


larchunix:
EncFs allows writing to read-only files.

##### Environment #####

encfs-1.8.1
fuse-2.9.4
linux-4.1.6

##### Steps to reproduce #####

    touch ReadOnlyFile
    chmod 400 ReadOnlyFile
    echo write > ReadOnlyFile

The `echo` performs successfully with encfs

    cat ReadOnlyFile
    write

##### Expected results #####

The `echo` command should fail like it does with other filesystems

    bash: ReadOnlyFile: Permission denied

larchunix:
I ran `git bisect` and found 82ceb88 as the faulty commit.

When I "revert" the commit by using the option `-o default_permissions` with `encfs`, POSIX permissions are correctly respected.

rfjakob:
Ok, this is one of my commits.

Could you show an "ls -l" of the file and of the corresponding encrypted
file?
The easiest way to find the right encrypted file is to look at the
timestamps.

larchunix:
As requested:

- Unencrypted view

        $ ls -l
        -r-------- 1 csl csl 6 Oct 30 12:53 ReadOnlyFile

- Encrypted view

        $ ls -l 
        -r-------- 1 csl csl 22 Oct 30 12:53 JqivLg5GL3qZmpRdPliwowpg

rfjakob:
Do you run encfs as root? I don't see how that file can be writeable
otherwise.

rfjakob:
Okay, I can reproduce this. `open_readonly_workaround` kicks in, see https://github.com/vgough/encfs/blob/master/encfs/RawFileIO.cpp#L82 .

@vgough do you remember the purpose of this function? Does it still have one?

If not, I have patch ready for removing it at https://github.com/rfjakob/encfs/commit/a05fe4b4f1d5212fda979b71f826c87a1fad2ba6 . That fixes this issue.

larchunix:
I don't understand why the behaviour is different when `encfs` is executed with `-o default_permissions`. IIUC, the `open_readonly_workaround` function is called in both cases.

rfjakob:
With `-o default_permissions`, libfuse checks if the operation is supposed to be allowed even before encfs gets involved.

vgough:
I believe that this workaround was in place in order to support --public mode.  But it has been there over 10 years, so it could be long out of date.

@rfjakob can you check that your patch works with --public mount mode?

[...]

PR #130, opened 2015-11-30
Remove open_readonly_workaround

rfjakob:
Since commit 82ceb88998fccfdbe21d2e66cd8764adef7d55e5 has removed
libfuse permissions checking, this function has caused read-only
files to writeable.

There seems to be no valid use case for writing to a read-only file (anymore?),
remove the function.

Fixes issue #112.


vgough:
Thanks Jakob.

PR #148, opened 2016-03-21
Bugfix: Labels are backwards for "Block" and "Block32" encoding

jetwhiz:
Block32 should be used for case-insensitive systems (not Block)

### DIFF
--- a/encfs/FileUtils.cpp
+++ b/encfs/FileUtils.cpp
@@ -1016,7 +1016,14 @@ RootPtr createV6Config(EncFS_Context *ctx, const shared_ptr<EncFS_Opts> &opts) {
     keySize = 256;
     blockSize = DefaultBlockSize;
     alg = findCipherAlgorithm("AES", keySize);
-    nameIOIface = BlockNameIO::CurrentInterface();
+
+	// If case-insensitive system, opt for Block32 filename encoding 
+#if defined(__APPLE__) || defined(WIN32)
+	nameIOIface = BlockNameIO::CurrentInterface(true);
+#else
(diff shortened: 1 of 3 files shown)


jetwhiz:
One more area I forgot: https://github.com/jetwhiz/encfs4win/commit/2fff5a3e4852225af8fbaba8aced2e80d4372701

jetwhiz:
I recommend we also choose Block32 as the default setting for Windows and Apple systems (for standard and paranoid mode) to prevent this from causing issues on these systems in the future.  

See related patch: https://github.com/jetwhiz/encfs4win/commit/4007a4e6393beb1819d7f986d5103521a92b4e43

encfs/FileUtils.cpp
```diff
@@ -1016,7 +1016,14 @@ RootPtr createV6Config(EncFS_Context *ctx, const shared_ptr<EncFS_Opts> &opts) {
     keySize = 256;
     blockSize = DefaultBlockSize;
     alg = findCipherAlgorithm("AES", keySize);
-    nameIOIface = BlockNameIO::CurrentInterface();
+
+	// If case-insensitive system, opt for Block32 filename encoding
```
vgough: formatting here and in the block below looks inconsistent with surrounding lines

vgough:
Thanks @jetwhiz , looks good.  Can you update the formatting in the last chunk of code?

jetwhiz:
Hi @vgough -- OK, I just pushed those changes into the branch.  It looks like I had used tabs instead of spaces for the improperly-formatted areas.

Issue #15, opened 2014-08-26
Possible Out of Bounds Write in StreamNameIO and BlockNameIO


lipnitsk:
From: https://defuse.ca/audits/encfs.htm

There is a possible buffer overflow in the encodeName method of [StreamNameIO](https://github.com/vgough/encfs/blob/master/encfs/StreamNameIO.cpp#L105) and [BlockNameIO](https://github.com/vgough/encfs/blob/master/encfs/BlockNameIO.cpp#L144). The methods write to the 'encodedName' argument without checking its length. This may allow an attacker with control over file names to crash EncFS or execute arbitrary code.

PR #149, opened 2016-03-23
Bugfix: Possible Out of Bounds Write in StreamNameIO and BlockNameIO #15

jetwhiz:
Issue #15, the encodeName functions fail to verify buffer length can store encoded filenames
For good measure and interface consistency, also check decodeName functions

### DIFF
--- a/encfs/BlockNameIO.cpp
+++ b/encfs/BlockNameIO.cpp
@@ -137,16 +137,15 @@ int BlockNameIO::maxDecodedNameLen(int encodedNameLen) const {
 int BlockNameIO::encodeName(const char *plaintextName, int length, uint64_t *iv,
                             char *encodedName, int bufferLength) const {
 
-  // copy the data into the encoding buffer..
-  rAssert(length <= (bufferLength - 2));
-  memcpy(encodedName + 2, plaintextName, length);
-
   // Pad encryption buffer to block boundary..
   int padding = _bs - length % _bs;
   if (padding == 0) padding = _bs;  // padding a full extra block!
 
-  rAssert(padding <= (bufferLength - length - 2));
+  rAssert(bufferLength >= length + 2 + padding);
   memset(encodedName + length + 2, (unsigned char)padding, padding);
+
+  // copy the data into the encoding buffer..
+  memcpy(encodedName + 2, plaintextName, length);
 
   // store the IV before it is modified by the MAC call.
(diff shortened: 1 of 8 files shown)


encfs/BlockNameIO.cpp
```diff
@@ -135,14 +135,17 @@ int BlockNameIO::maxDecodedNameLen(int encodedNameLen) const {
...
+
   // copy the data into the encoding buffer..
+  rAssert(length <= (bufferLength - 2));
   memcpy(encodedName + 2, plaintextName, length);
 
   // Pad encryption buffer to block boundary..
   int padding = _bs - length % _bs;
   if (padding == 0) padding = _bs;  // padding a full extra block!
 
+  rAssert(padding <= (bufferLength - length - 2));
```
vgough: this assert could be moved up to make the previous one unnecessary.

ie,  rAssert(bufferLength >= length + 2 + padding) eliminates the need for an rAssert(bufferLength >= length + 2).

encfs/StreamNameIO.cpp
```diff
@@ -104,11 +104,13 @@ int StreamNameIO::encodeName(const char *plaintextName, int length,
     encodedName[0] = (mac >> 8) & 0xff;
     encodedName[1] = (mac)&0xff;
     encodeBegin = (unsigned char *)encodedName + 2;
+	rAssert(length <= (bufferLength - 2));
   } else {
     // encfs 0.x stored checksums at the end.
     encodedName[length] = (mac >> 8) & 0xff;
     encodedName[length + 1] = (mac)&0xff;
     encodeBegin = (unsigned char *)encodedName;
+	rAssert(length <= bufferLength);
```
vgough: shouldn't this be the same as the previous assert? they both store 2 checksum bytes.

jetwhiz: You're right, I didn't notice we were writing to length and length+1.  I have fixed this and will push the changes.

encfs/BlockNameIO.cpp
jetwhiz: Since "padding" isn't defined until after the first assert, should I rearrange the writes so that we write the padding first, and then write the data?

jetwhiz:
Are we able to close issue https://github.com/vgough/encfs/issues/15 at this point?

Issue #81, opened 2015-05-25
Check password method


benrubson:
Hello,

Could it be possible to implement a method to check if a given password is good ?
For example as an option to encfsctl.
echo "pass" | encfsctl checkpasswd /my_encfs/
This would then return 0 or 1.

Thank you !

Ben

rfjakob:
What is wrong with mounting it?

benrubson:
I would like to use the password EncFS stores for other purpose than mounting the encrypted filesystem.
So I need to check the password I have against the one EncFS stores.

The following works to change password :
echo "oldpass
newpass" | encfsctl autopasswd /my_encfs/

So I assume that the following should not be difficult to implement :
echo "pass" | encfsctl checkpasswd /my_encfs/

Thank you very much !

benrubson:
Hello,

Could it be possible to write a quick Perl script which would check a password against a `.encfs6.xml` file ?

For example :
`check_encfs_password.pl <password> <.encfs6.xml_file>`
It would return 0 or 1 whether password is correct or not.

Could you indicate me the fields to use in the `.encfs6.xml` file and the operation to make on the password and these fields to be able to make the comparison ?

Thank you very much :+1: 

Ben

vgough:
Thanks Ben.

benrubson:
Thanks Valient !

PR #150, opened 2016-03-26
Add method to check password

benrubson:
Hello,

This PR adds a method to simply check password for volume, without changing it, without mounting the volume.

It solves #81.

Thank you very much :+1: 

Ben

### DIFF
--- a/encfs/encfsctl.cpp
+++ b/encfs/encfsctl.cpp
@@ -59,6 +59,7 @@ static int showInfo(int argc, char **argv);
 static int showVersion(int argc, char **argv);
 static int chpasswd(int argc, char **argv);
 static int chpasswdAutomaticly(int argc, char **argv);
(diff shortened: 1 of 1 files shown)


encfs/encfsctl.cpp
```diff
@@ -59,6 +59,7 @@ static int showInfo(int argc, char **argv);
 static int showVersion(int argc, char **argv);
 static int chpasswd(int argc, char **argv);
 static int chpasswdAutomaticly(int argc, char **argv);
+static int ckpasswdAutomaticly(int argc, char **argv);
```
vgough: Looks like this PR is out of date, as it duplicates changes already in master.

benrubson: No, these are 2 differences methods, `chpasswdAutomaticly`to change, `ckpasswdAutomaticly`to only check 👍

PR #165, opened 2016-05-10
Filenode cache refactor

ial0:
I've been working on a number of code refactoring's, this one is quite small and un-intrusive.

This simplifies the FileNode open file cache, removing PlaceHolder and the use of std::set.

The use of std::set here was redundant, the PlaceHolder pointer added was always new, guaranteeing its uniqueness.

PlaceHolder itself is an additional container which is unnecessary, simply store the shared_ptr in a list and pass the FileNode pointer directly to fuse


vgough:
Thanks!

PR #167, opened 2016-05-11
Add new command line options

jetwhiz:
Add new command line options

--mount  - only mount FS (do not create)
--unmount  - unmount FS
--config=path  - specify config file (override ENV)
~~--password=passwd  - specify password~~

### DIFF
--- a/encfs/FileUtils.cpp
+++ b/encfs/FileUtils.cpp
@@ -231,9 +231,20 @@ ConfigType readConfig_load(ConfigInfo *nm, const char *path,
  * Try to locate the config file
  * Tries the most recent format first, then looks for older versions
  */
-ConfigType readConfig(const string &rootDir, EncFSConfig *config) {
+ConfigType readConfig(const string &rootDir, EncFSConfig *config, 
+                      const string &cmdConfig) {
   ConfigInfo *nm = ConfigFileMapping;
   while (nm->fileName) {
+    // allow command line argument to override default config path 
+    if (!cmdConfig.empty()) {
+      if (!fileExists(cmdConfig.c_str())) {
+        RLOG(ERROR)
+            << "fatal: config file specified on command line does not exist: "
+            << cmdConfig;
+        exit(1);
+      }
+      return readConfig_load(nm, cmdConfig.c_str(), config);
+    }
     // allow environment variable to override default config path
     if (nm->environmentOverride != NULL) {
       char *envFile = getenv(nm->environmentOverride);
@@ -441,14 +452,18 @@ bool readV4Config(const char *configFile, EncFSConfig *config,
 }
 
 bool saveConfig(ConfigType type, const string &rootDir,
-                const EncFSConfig *config) {
+                const EncFSConfig *config, const string &cmdConfig) {
   bool ok = false;
 
   ConfigInfo *nm = ConfigFileMapping;
   while (nm->fileName) {
     if (nm->type == type && nm->saveFunc) {
       string path = rootDir + nm->fileName;
-      if (nm->environmentOverride != NULL) {
+      if (!cmdConfig.empty()) {
+        // use command line argument if specified
+        path.assign(cmdConfig);
+      }
+      else if (nm->environmentOverride != NULL) {
         // use environment file if specified..
         const char *envFile = getenv(nm->environmentOverride);
         if (envFile != NULL) path.assign(envFile);
@@ -1182,7 +1197,7 @@ RootPtr createV6Config(EncFS_Context *ctx,
     return rootInfo;
   }
 
-  if (!saveConfig(Config_V6, rootDir, config.get())) {
+  if (!saveConfig(Config_V6, rootDir, config.get(), opts->config)) {
     return rootInfo;
   }
 
@@ -1529,7 +1544,7 @@ RootPtr initFS(EncFS_Context *ctx, const std::shared_ptr<EncFS_Opts> &opts) {
   RootPtr rootInfo;
   std::shared_ptr<EncFSConfig> config(new EncFSConfig);
 
-  if (readConfig(opts->rootDir, config.get()) != Config_None) {
+  if (readConfig(opts->rootDir, config.get(), opts->config) != Config_None) {
     if (config->blockMACBytes == 0 && opts->requireMac) {
       cout << _(
           "The configuration disabled MAC, but you passed --require-macs\n");
--- a/encfs/FileUtils.h
+++ b/encfs/FileUtils.h
@@ -30,6 +30,16 @@
 #include "Interface.h"
 #include "encfs.h"
 
+
+ /* Arbitrary identifiers for long options that do
+ * not have a short version */
+#define LONG_OPT_ANNOTATE 513
+#define LONG_OPT_NOCACHE 514
+#define LONG_OPT_REQUIRE_MAC 515
+#define LONG_OPT_MOUNT 516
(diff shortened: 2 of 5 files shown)


encfs/FileUtils.cpp
```diff
@@ -1160,10 +1171,10 @@ RootPtr createV6Config(EncFS_Context *ctx,
 
   // get user key and use it to encode volume key
   CipherKey userKey;
-  VLOG(1) << "useStdin: " << useStdin;
-  if (useStdin) {
+  VLOG(1) << "passSrc: " << opts->passSrc;
+  if (opts->passSrc != Pass_Prompt) {
```
vgough: shouldn't this be   opts->passSrc == Pass_Stdin   ?

jetwhiz: You're right -- on Windows we don't have `passwordProgram`, so I didn't see that this will never be executed with the current layout on Linux.  

We want to fall into `userKey = config->getUserKey(opts);` for either Pass_Stdin or Pass_Cmd.  What do you think about changing this if-statement to: 

```C++
if (opts->passSrc == Pass_Cmd || opts->passSrc == Pass_Stdin) {
```

gh223682:
@vgough @jetwhiz I don't think that having an option to specify the password from the command-line is a good idea due to process security issues. This shouldn't be in the core code.

jetwhiz:
TrueCrypt/VeraCrypt both allow for passing the password in via command line options (see https://veracrypt.codeplex.com/wikipage?title=Command%20Line%20Usage).  

It can be useful when you want to create encrypted filesystems in an automated fashion in software (e.g., encfs --standard --password="$usrpass" /home/me /tmp/crypt-view) without needing to create pipes to the program to send the password to it.  Usage of it should be discouraged (in documentation) on the command line, since it may maintain a history, though.

Aikhjarto:
> Usage of it should be discouraged (in documentation) on the command line, since it may maintain a history, though.

Not only history of the shell is a problem. Under Linux the command `ps aux` shows the command line of all running processes. This will exhibit the password while encfs is running.

pepa65:
For scripting, the option to pass some form of password to the binary is essential, either through a string, a pipe, an external password utility, or any means.

gh223682:
> TrueCrypt/VeraCrypt both allow for passing the password in via command line options

@jetwhiz Hmm, that's interesting. There are other programs like fetchmail which don't have an option like this explicitly for process security reasons.

> It can be useful when you want to create encrypted filesystems in an automated fashion in software (e.g., encfs --standard --password="$usrpass" /home/me /tmp/crypt-view) without needing to create pipes to the program to send the password to it. Usage of it should be discouraged (in documentation) on the command line, since it may maintain a history, though. 

I was going to propose an alternate solution for people who want to launch EncFS with an unattended script (which includes me): an option to specify a permissions-restricted file which contains the password, e.g. "--password-file".

rfjakob:
This option already exists through "--extpass". The difference to
"--password" is that you can use it securely using something like "cat
mypassword.txt".

gh223682:
@rfjakob For the --extpass= option, an executable binary file is required, even if it is just to cat a file.

On certain platforms like Windows, there is no default system binary which can do this. Commands like "echo" and "type" are built-in cmd.exe shell commands.

Aikhjarto:
> Commands like "echo" and "type" are built-in cmd.exe shell commands and do not work.

Why not invoking cmd.exe with something like `cmd /c type mypassword.txt`?

gh223682:
@Aikhjarto That'll work.

Mind you, I'm just trying to suggest an acceptable/equivalent alternative to a "--password=" option.

If there are no issues implementing a "--password=" option, then there's no point for a "--password-file=" option.

jetwhiz:
I think ``cmd /c type mypassword.txt`` is a decent enough solution coupled with --extpass.  If you guys prefer, I can remove the "--password=passwd" portion of this PR, unless this functionality is helpful to someone else.

rfjakob:
Yes please, there is no way to use `--password` securely.

jetwhiz:
The latest version should now include: 

--mount - only mount FS (do not create)
--unmount - unmount FS
--config=path - specify config file (override ENV)

imperative:
Guys, we should really allow using the password in command line.
Encfs is a solution that is already geared towards sophisticated users. We should trust the users to know the implications of this. Especially when we put a warning in the documentation.

At the same time, there are many legitimate use cases for this. 
For example when using encfs to reverse-encrypt backups in the cloud. In that case we want the mounting to be automatic and we don't care if the password is easily readable locally, we only care that the cloud gets encrypted files.

pepa65:
Hear hear! Please don't take essential options away.

rfjakob:
Something like `--passfile` would be acceptable.

rfjakob:
@jetwhiz Looks good to me, but can you please squash this into a single commit?
@vgough Ok to merge?

pepa65:
I do hope the --stdinpass option stays, but I don't understand why you would remove options that users want, have a good use case for, and don't take a lot of code.

rfjakob:
Dear @pepa65 , no options will be removed, this discussion is strictly about adding new ones!

jetwhiz:
@rfjakob Should I open a new pull request with the squashed commits, or is there a way to do that on GitHub?

rfjakob:
Yes, just "git push --force" into the branch that you used for the pull
request. The changed commits will show up automatically in the pull request.

Zero3:
Good idea with these new command-line options @jetwhiz! I've been missing the mount/unmount commands.

I'd like to add some thoughts about passing passwords through files, though. This technique does not appear to be secure. I see a number of potential issues:

1) Even if the file is deleted from the file system, the file contents may be recoverable. As far as I know, recovery is pretty trivial on spinning disks unless contents are overwritten before file deletion. On SSDs, you cannot reliably overwrite file contents as the disk might save the new contents to different blocks instead of overwriting. Recovery is probably more difficult though.

2) The user might have backup software running that automatically backs up the password file.

3) OS-level file caching and swap might pose risks too.

One could probably work around these, but I'm a little concerned about adding features that let users use their passwords in insecure ways. If such features are added, I hope you add some big fat warning messages next to them.

imperative:
IMO there should be warnings, but the options should definitely be there. Otherwise it's very hard to automate the operation, which is an essential functionality for some users. 

If these options are not implemented, it will force users to either use other encryption software (which leads to segmentation of the field and ultimately more code which is spread out, which has more bugs and thus overall less security), or implement some external addons (like sshpass) which will introduce all the problems we discuss here, and even more, because those addons are not very likely to be implemented securely.

This is opensource, we should give users the most flexibility with decent defaults, and hide dangerous options behind warnings and manuals. If they wanted a dull solution they would use bitlocker.

Zero3:
@imperative I completely agree that it should be possible to integrate/automate encfs! I'm working on a project about this myself right now :).

I don't mind adding this feature, I'm just a little concerned, that's all. I don't know what the best practice for doing this kind of thing would be, it's just that passing the password through command line or file on disk seems somewhat insecure. Maybe there is a better way?

jetwhiz:
Hi @rfjakob @vgough  -- is there any update on merging these new options into encfs?

benrubson:
I am also interested in the `--config` option 👍 
@vgough, could we merge it if everything looks good to you ? (and release 1.9.2 ?)
Many thanks 👍

benrubson:
please see #474 for a rework 👍

PR #172, opened 2016-05-23
Fix incorrect masking in encfsctl export's copyContents

jetwhiz:
Code did not properly detect when a file had S_IFLNK mode


rfjakob:
Ack.

Issue #181, opened 2016-06-23
permission issue when having a git rep on a mounted encfs directory


disteph:
Using encfs 1.9.0.
I just set up an encfs directory (config file attached, I don't want to encode filenames)
[encfs6.xml.txt](https://github.com/vgough/encfs/files/329379/encfs6.xml.txt)

After I cd to the mount point, I do
$ git init
Initialised empty Git repository in [...]/.git/
$ echo "test" >> test
$ git add test
error: insufficient permission for adding an object to repository database .git/objects
error: test: failed to insert into database
error: unable to index file test
fatal: adding files failed

If I do it on the original directory, or any directory, of course it works. Just wondering what kind of weird operation git is trying to do that requires something that encfs is not providing. Any idea?
Can somebody reproduce this?

enodata:
Could not reproduce this on FreeBSD 10.3-STABLE, encfs 1.8.1. Might be smth introduced after 1.8.1?

rfjakob:
I can reproduce this with latest master.
strace:

    open(".git/objects/9d/tmp_obj_jL6IBf", O_RDWR|O_CREAT|O_EXCL, 0444) = -1 EACCES (Permission denied)

debug log:

    unique: 249, opcode: CREATE (35), nodeid: 13, insize: 71, pid: 23191
    create flags: 0x80c2 /.git/objects/9d/tmp_obj_DZOUJ0 0100444 umask=0002
       unique: 249, error: -13 (Permission denied), outsize: 16

rfjakob:
Reverting c2e046b694a42a6e17ec4a47a517638a098f928a fixes this.

rfjakob:
Well, that's pretty bad.

What git is doing:

1) open tmp_obj_DZOUJ0 (creates the file with 444 permissions but git gets a writeable file descriptor)

What encfs is doing:

1) open tmp_obj_DZOUJ0 (creates the file with 444 permissions)
2) close the file descriptor
3) open tmp_obj_DZOUJ0 again (fails because of 444 permissions)

open_readonly_workaround papered over this problem, but caused read-only files to be writeable ( bug #112 ). Unconditionally adding `-o default_permissions` masked that issue, but that was removed in https://github.com/vgough/encfs/commit/82ceb88998fccfdbe21d2e66cd8764adef7d55e5 .

I guess the only way to fix this without major surgery in EncFS is reverting both c2e046b and 82ceb88998fc .

disteph:
Ah, right.
[...]

PR #182, opened 2016-06-25
Fix issue 181 by reverting the responsible commits

rfjakob:
Also add a test case equivalent to

    umask 0777; echo foo > bar

to make sure we don't regress again.

Fixes issue #181 .


vgough:
Thanks for the test case Jakob.

samrocketman:
The branch should probably be deleted.  While I _can delete it_ I try to resist modifying state in this project and someone might want to keep said branch.

rfjakob:
I think deleting it is fine. Done.

PR #195, opened 2016-08-12
Bugfix: Segfault when running paranoid config in verbose mode

jetwhiz:
When "external IV chaining" is enabled, cipherName_ parameter may be NULL, resulting in segfault in verbose log attempt


rfjakob:
I'm sure the patch fixes the crash, but the actual bug is that cipherName_ is null, no?

jetwhiz:
@rfjakob  -- from what I understand, it is normal for this to be null for external IV chaining during a call to findOrCreate: 

https://github.com/vgough/encfs/blob/b0ef1d719eceb27f231ea9c467c63dbff78087b0/encfs/DirNode.cpp#L642

rfjakob:
I see. I think @vgough will agree that this is good to merge. Thanks!

PR #210, opened 2016-09-12
easylogging++ library available in newer version 9.84

svetlemodry:
no API change since v9.80

### DIFF
--- a/internal/easylogging++.h
+++ b/internal/easylogging++.h
@@ -1,8 +1,10 @@
 //
-//  Easylogging++ v9.80
+//  Bismillah ar-Rahmaan ar-Raheem
+//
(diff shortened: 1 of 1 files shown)


svetlemodry:
Hi!
I am the [encfs](https://www.archlinux.org/packages/community/x86_64/encfs/) package maintainer for Arch Linux. There seems to be a way to use the [system wide](https://git.archlinux.org/svntogit/community.git/tree/trunk/PKGBUILD?h=packages/encfs) tinyxml2 library for building encfs. I believe it would be good to have a similar possibility available for easylogging++ library.

tinyxml2 is also available as [v4.0.1](https://github.com/leethomason/tinyxml2/releases/tag/4.0.1) since mid July.

PR #222, opened 2016-10-03
CMakeList uses CMAKE_CURRENT_LIST_DIR

thejinx0r:
When including EncFS as a submodule in another project, encfs now use the paths relative to it self, rather than the parents path.


vgough:
Thanks!

Issue #200, opened 2016-08-28
Silence security-related warning in encfs/test.cpp


mouse07410:
```
. . . . .
[ 92%] Building CXX object CMakeFiles/checkops.dir/encfs/test.cpp.o
/Users/uri/src/encfs/encfs/test.cpp:178:22: warning: 'tmpnam' is deprecated: This function is provided for
      compatibility reasons only. Due to security concerns inherent in the design of tmpnam(3), it is highly
      recommended that you use mkstemp(3) instead. [-Wdeprecated-declarations]
    auto name = std::tmpnam(nullptr);
                     ^
/usr/include/stdio.h:276:7: note: 'tmpnam' has been explicitly marked deprecated here
char    *tmpnam(char *);
         ^
1 warning generated.
. . . . .
```

PR #228, opened 2016-10-15
silence security-related warning in encfs/test.cpp

rogeliodh:
Hi, I was working in issue #200 but I see that it is already fixed by using mkstemp. Anyway here it is my version using std::tmpfile() which has the advantage of not requiring to unlink the temporary file and that it is more standard (and cross platform) than mkstemp.

### DIFF
--- a/encfs/FileUtils.cpp
+++ b/encfs/FileUtils.cpp
@@ -522,8 +522,8 @@ tinyxml2::XMLElement *addEl<>(tinyxml2::XMLDocument &doc,
   return addEl(doc, parent, name, v.c_str());
 }
 
-tinyxml2::XMLDocument *writeV6Config(const EncFSConfig *cfg) {
-  tinyxml2::XMLDocument *docptr = new tinyxml2::XMLDocument();
+std::shared_ptr<tinyxml2::XMLDocument> writeV6Config(const EncFSConfig *cfg) {
+  std::shared_ptr<tinyxml2::XMLDocument> docptr (new tinyxml2::XMLDocument());
   tinyxml2::XMLDocument &doc = *docptr;
 
   // Various static tags are included to make the output compatible with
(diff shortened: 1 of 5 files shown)


encfs/FileUtils.cpp
```diff
@@ -263,22 +263,14 @@ ConfigType readConfig(const string &rootDir, EncFSConfig *config) {
...
-  XmlValuePtr serialization = rdr["boost_serialization"];
+bool readV6Config(XmlReader *rdr, EncFSConfig *cfg, struct ConfigInfo *) {
+  XmlValuePtr serialization = (*rdr)["boost_serialization"];
   XmlValuePtr config = (*serialization)["cfg"];
   if (!config) {
     config = (*serialization)["config"];
   }
   if (!config) {
-    RLOG(ERROR) << "Unable to find XML configuration in file " << configFile;
+    RLOG(ERROR) << "Unable to find XML configuration in config file ";;
```
samrocketman: Two semicolons.

rfjakob:
Hmm, the description of https://github.com/vgough/encfs/pull/228/commits/fa17097260d5aa120a01c4f88b82680006000e74 does not seem to match what the patch actually does. It seems to rewrite half of the config handling code, with the result being some 22 lines larger than what we started with.

To be honest, at the first look I like the mkstemp version better.

rogeliodh:
Yes, it rewrites the config handling code to be able to use FILE pointers in addition to file descriptors (to be able use tmpfile()).

But yes, this is just more code. Although it is more standard and cross-platform, I'm not sure if it is valuable for this project.

I just wanted to share my already done work. It is up to you!

rogeliodh:
@ial0 it returns a pointer because XMLDocument does not support copy operation, so returning it by value is impossible. 

I've update the patch to use unique_ptr

benrubson:
I propose to close this for the moment, as another solution has already been merged, and as it's initially only related to `test.cpp`.
Of course we will re-open it if needed ! Thx 👍

Issue #179, opened 2016-06-11
build error with openssl 1.1


Code7R:
Greetings,
encfs won't compile with openssl 1.1 anymore. See https://breakpoint.cc/openssl-1.1-rebuild-2016-05-29/Attempted/encfs_1.8.1-3_amd64-20160529-1416 for build log and https://wiki.openssl.org/index.php/1.1_API_Changes for further information.

vgough:
Thanks, change merged.

PR #230, opened 2016-10-15
build with openssl 1.1.0b

rogeliodh:
Hi, I've started working in building encfs with openssl 1.1 (#179), this is my first attempt.

It builds and seems to work, But this needs work yet:

- There are still some deprecation warnings 
- Make it backward compatible with openssl < 1.1

### DIFF
--- a/encfs/SSL_Cipher.cpp
+++ b/encfs/SSL_Cipher.cpp
@@ -74,19 +74,19 @@ int BytesToKey(int keyLen, int ivLen, const EVP_MD *md,
   int nkey = key ? keyLen : 0;
   int niv = iv ? ivLen : 0;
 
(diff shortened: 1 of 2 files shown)


rfjakob:
Looks good, thanks! What breaks when compiling with older openssl?

rogeliodh:
Some functions were missing in OpenSSL < 1.1. I have now added a compat header, it builds with OpenSSL 1.0 and OpenSSL 1.1 now.

Code7R:
Is there a way to get an official release with openssl 1.1 support?

Debian Sid is basically going to deprecate older openssl very soon, and I would like to package it as an official release and not with a patch picked from master branch.

PR #235, opened 2016-10-23
remove usage of deprecated RAND_pseudo_bytes

rogeliodh:
OpenSSL 1.1 deprecated RAND_pseudo_bytes because RAND_bytes behave practically the same, except that "one will add an error to the error queue with insufficient entropy whilst the other will not" [1]. And we already manage the error queue properly, so we can remove it.

This commit removes a deprecated warning during building with OpenSSL 1.1

[1] https://git.openssl.org/gitweb/?p=openssl.git;a=commitdiff;h=302d38e3f73d5fd2ba2fd30bb7798778cb9f18dd

### DIFF
--- a/encfs/SSL_Cipher.cpp
+++ b/encfs/SSL_Cipher.cpp
@@ -488,19 +488,15 @@ static uint64_t _checksum_64(SSLKey *key, const unsigned char *data,
 /**
  * Write "len" bytes of random data into "buf"
  *
- * See "man 3 RAND_bytes" for the effect of strongRandom
+ * We ignore the @strongRandom parameter because OpenSSL
+ * does not * offer a "weak" random generator
  */
 bool SSL_Cipher::randomize(unsigned char *buf, int len,
-                           bool strongRandom) const {
+                           bool /*strongRandom*/) const {
   // to avoid warnings of uninitialized data from valgrind
   memset(buf, 0, len);
-  int result;
-  if (strongRandom) {
-    result = RAND_bytes(buf, len);
-  } else {
-    result = RAND_pseudo_bytes(buf, len);
-  }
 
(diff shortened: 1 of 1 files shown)


vgough:
The distinction between strong and weak randomness is an important one in practice.  Where we can live with weak randomness, we can prioritize availability over security and use whatever randomness is available.  Where we need strong randomness, it is more important than availability of the application - eg we would rather not create a new filesystem than create one with a potentially weak key.

Even if OpenSSL is removing this distinction, which at face value seems like a bad idea, I'd rather that the encfs code clearly indicate its own requirements.  If nothing else, that'll make it easier to move to a different source of randomness.

rogeliodh:
OpenSSL is not removing the distinction, they simply are removing the RAND_pseudo_bytes because it really doesn't return pseudo random data, it returns the same random data as RAND_bytes (you can see it here so https://github.com/openssl/openssl/blob/master/crypto/rand/md_rand.c#L504).

So, to keep the strongRandom parameter, encfs would need to use another function when strongRandom==false. I don't know what function we could use instead (OpenSSL doesn't seem to provide one). A temporary alternative could be to keep the strongRandom parameter but ignore it in SSL_Cipher::randomize (and always call  RAND_bytes).

vgough:
Thanks for the explanation.  I'd rather not remove the strongRandom argument, since it provides a bit of information that can be useful in other branches where I use alternatives to OpenSSL.  Replacing just the switch between functions in SSL_Cipher.cpp seems sufficient.

rogeliodh:
Hi, I've updated the pull request just ignoring the strongRandom in SSL_Cipher.cpp as @vgough suggests.

vgough:
Thanks @rogeliodh .

Issue #241, opened 2016-10-28
segfault when running encfs with -S for the first time


pschiffe:
```
$ rpm -q fuse-encfs
fuse-encfs-1.9.1-2.fc24.x86_64
```

I would like to script the creation of encrypted folder with encfs. So I've tried:

```
$ mkdir a b
$ echo 'qwerty' | encfs -S $(pwd)/a $(pwd)/b
Creating new encrypted volume.
Please choose from one of the following options:
 enter "x" for expert configuration mode,
 enter "p" for pre-configured paranoia mode,
 anything else, or an empty line will select standard mode.
?> 
Standard configuration selected.

Configuration finished.  The filesystem to be created has
the following properties:
Filesystem cipher: "ssl/aes", version 3:0:2
Filename encoding: "nameio/block", version 4:0:2
Key Size: 192 bits
Block Size: 1024 bytes
Each file contains 8 byte header with unique IV data.
Filenames encoded using IV chaining mode.
File holes passed through to ciphertext.

Now you will need to enter a password for your filesystem.
You will need to remember this password, as there is absolutely
no recovery mechanism.  However, the password can be changed
later using encfsctl.

Zero length password not allowed
[1]    31157 done                              echo 'qwerty' | 
       31160 segmentation fault (core dumped)  encfs -S $(pwd)/a $(pwd)/b
```

I'm not sure whether it would work without that seg fault, but that shouldn't occur in any way. Can you reproduce it? Also, is it possible to script creation of encrypted directory without using `expect` utility? Default options are fine with me.

samrocketman:
I've not run into that.  I'll try reproducing it.  For now, check out my script.

https://github.com/samrocketman/home/blob/master/bin/mount-encfs-share.sh

samrocketman:
My script also has an example to set the password via stdin.

pschiffe:
Thanks for your link, I'll check it out. I'm attaching archive with all abrt data about the crash (also containing coredump). Hopefully it will help to investigate the problem [encfscrash.zip](https://github.com/vgough/encfs/files/560778/encfscrash.zip)

pschiffe:
I've created bug report in Fedora bugzilla and we've found a workaround:
`echo -e '\nqwerty' | encfs -S $(pwd)/a $(pwd)/b`

https://bugzilla.redhat.com/show_bug.cgi?id=1390107

Vascom:
> My script also has an example to set the password via stdin.

[...]

PR #243, opened 2016-10-31
Fix a segfault when password is zero length.

ial0:
This patch fixes the issues in #241.

* Added a missing `nullptr` check on `userKey`

* If `useStdin` and `configMode == Config_Prompt` then default to `Config_Standard`,
   otherwise we might read the password input instead on the config mode.

I'm not sure if setting `Config_Standard` as a default option like this is the best course of action.


vgough:
Looks good, thanks!

Issue #50, opened 2015-01-10
symbol lookup error


pepa65:
I have pulled the latest master, configure/made it, and now I get:
undefined symbol: _Z6initFSP13EncFS_ContextRKSt10shared_ptrI10EncFS_OptsE

pepa65:
Started from scratch, cloned current master, prepped for Ubuntu, but the configure step now ends with:
cannot find input file: `po/Makefile.in.in'

rfjakob:
Which ubuntu version?

pepa65:
14.04.1

rfjakob:
Could you copy-paste the whole shell output from "git clone" to the error?

pepa65:
$ git clone https://github.com/vgough/encfs
Cloning into 'encfs'...
remote: Counting objects: 1997, done.
remote: Total 1997 (delta 0), reused 0 (delta 0)
Receiving objects: 100% (1997/1997), 1021.05 KiB | 93.00 KiB/s, done.
Resolving deltas: 100% (1434/1434), done.
Checking connectivity... done.
$ libtoolize --force
libtoolize: putting auxiliary files in AC_CONFIG_AUX_DIR, `build-aux'.
libtoolize: linking file `build-aux/ltmain.sh'
libtoolize: putting macros in AC_CONFIG_MACRO_DIR, `m4'.
libtoolize: linking file `m4/libtool.m4'
libtoolize: linking file `m4/ltoptions.m4'
libtoolize: linking file `m4/ltsugar.m4'
libtoolize: linking file `m4/ltversion.m4'
libtoolize: linking file `m4/lt~obsolete.m4'
$ aclocal
$ autoheader
$ touch build-aux/config.rpath
$ automake --force-missing --add-missing
configure.ac:29: installing 'build-aux/compile'
configure.ac:9: installing 'build-aux/config.guess'
configure.ac:9: installing 'build-aux/config.sub'
configure.ac:7: installing 'build-aux/install-sh'
configure.ac:7: installing 'build-aux/missing'
encfs/Makefile.am: installing 'build-aux/depcomp'
$ autoconf
$ ./configure --with-boost-libdir=/usr/lib/x86_64-linux-gnu
checking for a BSD-compatible install... /usr/bin/install -c
checking whether build environment is sane... yes
checking for a thread-safe mkdir -p... /bin/mkdir -p
checking for gawk... gawk
checking whether make sets $(MAKE)... yes
checking whether make supports nested variables... yes
[...]

PR #258, opened 2016-12-23
Fix up encfssh

charles-dyfis-net:
The original implementation of encfssh had several issues:

- Could not safely handle directory names with spaces
- Could not safely handle paths starting with a dash (`-`)
- Option strings could not be passed with spaces
- Could print incorrect help based on filenames in current directory
- Used `test` syntax deprecated in current POSIX.2 standard
- Had shell injection vulnerabilities based on both current working directory and mount-point directory names.
- Failures of the canonicalize function could silently return the current directory rather than the target.
- Evaluated boolean variables as commands -- requiring audit re: whether any codepaths could leave these uninitialized, allowing arbitrary commands to be run. To avoid any need for such audit, use explicit string comparisons to determine booleans' values.

In this version, shell injection is only permitted through the (explicitly defined) `FUSE_UMOUNT` mechanism.

This version also adds support for the GNU-style "--help" argument.

### DIFF
--- a/encfs/encfssh
+++ b/encfs/encfssh
@@ -62,9 +62,13 @@ echo "Directory is $unenc_dir" >&2
 cd "$unenc_dir" || exit
 
 # Fall back to umount if fusermount is not available (e.g., on OS X)
-FUSE_UMOUNT="$(command -v fusermount)"
-FUSE_UMOUNT="${FUSE_UMOUNT:+fusermount -u}"
-FUSE_UMOUNT="${FUSE_UMOUNT:-umount}"
+fuse_umount() {
+	if command -v fusermount >/dev/null 2>&1; then
+		fusermount -u "$@"
+	else
+		umount "$@" # MacOS case
+	fi
+}
 
 # Honor the SHELL environment variable to select a shell to run
 "$SHELL"; retval=$?
@@ -74,17 +78,9 @@ FUSE_UMOUNT="${FUSE_UMOUNT:-umount}"
 #    failing to check exit status from cd.
 cd / ||:
 
-# Note that there are two ways FUSE_UMOUNT can be parsed:
-# - With eval (process command substitutions, redirections, etc., and honors
-#   quotes; should only be done with human-generated/reviewed/trusted content,
-#   see [BashFAQ #48](http://mywiki.wooledge.org/BashFAQ/048))
-# - Without (can't run all possible commands this way, see
-#   [BashFAQ #50](http://mywiki.wooledge.org/BashFAQ/050)).
-#
-# Since the expectation is that this will be user-controlled, flexibility is
-# presumably more important than security, so opting for the eval approach.
+# if unmount fails, skip rmdir, always use exit status of failure
+fuse_umount "$unenc_dir" || exit
 
-eval "$FUSE_UMOUNT \"\$unenc_dir\"" # Call hook, honoring any shell syntax/quotes/&c. in same
 if ! [ "$unenc_dir_given" = true ]; then
 	rmdir -- "$unenc_dir"
 fi
--- a/encfs/encfssh
+++ b/encfs/encfssh
@@ -58,7 +58,7 @@ fi
 if ! [ "$unenc_dir_given" = "true" ]; then
     chmod 700 "$unenc_dir"
 fi
-echo "Directory is $unenc_dir"
+echo "Directory is $unenc_dir" >&2
 cd "$unenc_dir" || exit
 
 # Fall back to umount if fusermount is not available (e.g., on OS X)
@@ -66,13 +66,14 @@ FUSE_UMOUNT="$(command -v fusermount)"
 FUSE_UMOUNT="${FUSE_UMOUNT:+fusermount -u}"
 FUSE_UMOUNT="${FUSE_UMOUNT:-umount}"
 
-# Set the shell up
-#
-# Arguments are passed out-of-band to make all eval invocations explicit.
-# (Using double quotes to generate the script passed to sh -c can lead to
-# unintentional eval of content intended to be treated as data rather than code
-# -- consider for example the case where `unenc_dir=$'/tmp/evil\n/$(rm -rf $HOME)\'$(rm -rf $HOME)\'/')`.
-#
+# Honor the SHELL environment variable to select a shell to run
+"$SHELL"; retval=$?
+
+# ensure that this shell isn't itself holding the mounted directory open
+# ...but avoid terminating on failure, *or* causing a shellcheck warning for
+#    failing to check exit status from cd.
+cd / ||:
+
 # Note that there are two ways FUSE_UMOUNT can be parsed:
 # - With eval (process command substitutions, redirections, etc., and honors
 #   quotes; should only be done with human-generated/reviewed/trusted content,
@@ -83,20 +84,8 @@ FUSE_UMOUNT="${FUSE_UMOUNT:-umount}"
 # Since the expectation is that this will be user-controlled, flexibility is
 # presumably more important than security, so opting for the eval approach.
 
-# Q: Why are we handing off to /bin/sh at all, rather than doing this from here?
-
-# Note: all-caps variable names may be reserved; use lower-case names
-# exclusively in child shell.  Avoids breakage if SHELL is read-only, f/e.
-# See http://pubs.opengroup.org/onlinepubs/009695399/basedefs/xbd_chap08.html
-# specifying (only!) lower-case names as guaranteed to be available for
-# application use.
-exec /bin/sh -c '
-fuse_umount=$1; shell=$2; unenc_dir=$3; unenc_dir_given=$4
-"$shell"; retval=$?
-cd /
-eval "$fuse_umount \"\$unenc_dir\""
+eval "$FUSE_UMOUNT \"\$unenc_dir\"" # Call hook, honoring any shell syntax/quotes/&c. in same
 if ! [ "$unenc_dir_given" = true ]; then
 	rmdir -- "$unenc_dir"
 fi
 exit "$retval"
-' _ "$FUSE_UMOUNT" "$SHELL" "$unenc_dir" "$unenc_dir_given"
--- a/encfs/encfssh
+++ b/encfs/encfssh
@@ -6,16 +6,17 @@
 # Contributed by David Rosenstrauch.
 
 canonicalize() {
-	cd "$1"
+	cd "$1" || return
 	pwd
 }
 
 
-if [ -z "$1" -o "$1" = "-h" ]; then
-	echo Usage: encfssh encrypted_directory [unencrypted-directory [-p]]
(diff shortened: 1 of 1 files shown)


charles-dyfis-net:
(Oh -- a change I missed in the descriptive text: `which` is not a POSIX-specified command, nor a shell builtin. `command -v`, by contrast, is [guaranteed to work](http://pubs.opengroup.org/onlinepubs/009696799/utilities/command.html) on all shells with the (admittedly optional) POSIX user portability extension; it's thus a builtin even with ash, dash, etc).

rfjakob:
Good improvements!

Maybe I am missing something, but why does the unmounting have to happen in a subshell?

charles-dyfis-net:
Heh. Color me embarrassed for retaining that behavior without cause.

encfs/encfssh
```diff
@@ -28,42 +29,63 @@ if [ ! -z "$2" ]; then
...
 
 enc_dir=$(canonicalize "$enc_dir")
 unenc_dir=$(canonicalize "$unenc_dir")
 
-options=""
-if $unenc_dir_given; then
-    if $mount_public; then
-	options="-- -o allow_other"
+# clear command-line argument list; hereafter used for options
+set --
```
rfjakob: I like the `options` variant better. It's simple and explicit in what it does, unlike the $@ trick, which I have personally never seen

encfs/encfssh
```diff
@@ -6,16 +6,17 @@
 # Contributed by David Rosenstrauch.
```
rfjakob: How about `#/bin/bash -eu` ?

encfs/encfssh
```diff
@@ -28,42 +29,63 @@ if [ ! -z "$2" ]; then
...
 # Fall back to umount if fusermount is not available (e.g., on OS X)
-FUSE_UMOUNT="$(which 2>/dev/null fusermount)"
+FUSE_UMOUNT="$(command -v fusermount)"
 FUSE_UMOUNT="${FUSE_UMOUNT:+fusermount -u}"
 FUSE_UMOUNT="${FUSE_UMOUNT:-umount}"
 
-# Set the shell up
-exec /bin/sh -c "$SHELL ; cd $orig_dir ; $FUSE_UMOUNT $unenc_dir ; if ! $unenc_dir_given; then rmdir $unenc_dir; fi"
+# Honor the SHELL environment variable to select a shell to run
+"$SHELL"; retval=$?
```
rfjakob: What does this do?

encfs/encfssh
```diff
@@ -28,42 +29,63 @@ if [ ! -z "$2" ]; then
...
-orig_dir=$(pwd)
-cd $unenc_dir
+echo "Directory is $unenc_dir" >&2
+cd "$unenc_dir" || exit
 
 # Fall back to umount if fusermount is not available (e.g., on OS X)
-FUSE_UMOUNT="$(which 2>/dev/null fusermount)"
+FUSE_UMOUNT="$(command -v fusermount)"
 FUSE_UMOUNT="${FUSE_UMOUNT:+fusermount -u}"
 FUSE_UMOUNT="${FUSE_UMOUNT:-umount}"
```
rfjakob: This should all be done *before* mounting the filesystem.

encfs/encfssh
```diff
@@ -28,42 +29,63 @@ if [ ! -z "$2" ]; then
...
 
-# Set the shell up
-exec /bin/sh -c "$SHELL ; cd $orig_dir ; $FUSE_UMOUNT $unenc_dir ; if ! $unenc_dir_given; then rmdir $unenc_dir; fi"
+# Honor the SHELL environment variable to select a shell to run
+"$SHELL"; retval=$?
+
+# ensure that this shell isn't itself holding the mounted directory open
+# ...but avoid terminating on failure, *or* causing a shellcheck warning for
+#    failing to check exit status from cd.
+cd / ||:
```
rfjakob: This should be done *before* mounting, and we SHOULD terminate on failure

encfs/encfssh
```diff
@@ -28,42 +29,63 @@ if [ ! -z "$2" ]; then
...
+# - With eval (process command substitutions, redirections, etc., and honors
+#   quotes; should only be done with human-generated/reviewed/trusted content,
+#   see [BashFAQ #48](http://mywiki.wooledge.org/BashFAQ/048))
+# - Without (can't run all possible commands this way, see
+#   [BashFAQ #50](http://mywiki.wooledge.org/BashFAQ/050)).
+#
+# Since the expectation is that this will be user-controlled, flexibility is
+# presumably more important than security, so opting for the eval approach.
+
+eval "$FUSE_UMOUNT \"\$unenc_dir\"" # Call hook, honoring any shell syntax/quotes/&c. in same
```
rfjakob: Why do we need `eval` here?

encfs/encfssh
charles-dyfis-net: Strongly opposed. See [BashFAQ #105](http://mywiki.wooledge.org/BashFAQ/105) and [BashFAQ #107](http://mywiki.wooledge.org/BashFAQ/107).

encfs/encfssh
charles-dyfis-net: It only *looks* simple -- the actual behavior of the options variant is considerably more complicated than how it reads. See [BashFAQ #50](http://mywiki.wooledge.org/BashFAQ/050) for an in-depth discussion of why strings are the wrong data structure to use for storing argument lists.

encfs/encfssh
charles-dyfis-net: No, we shouldn't: If we terminated now, we'd be leaving the filesystem mounted.

encfs/encfssh
charles-dyfis-net: Backwards compatibility. If we didn't use `eval`, then `FUSE_UNMOUNT` values which were previously supported (containing literal quotes intended to be parsed as syntactic) would no longer work.

encfs/encfssh
charles-dyfis-net: Gives the user their interactive shell. Could pass `-i` to make that even more explicit, but I don't want to rule out non-POSIX shells such as `fish` by assuming that POSIX conventions will hold.

encfs/encfssh
charles-dyfis-net: Why? The chain of commands can't possibly fail -- the fallback is just assigning the `umount` command, after all.

encfs/encfssh
rfjakob: Do we *need* the more complicated cases that the $@ construction handles?

encfs/encfssh
charles-dyfis-net: We could only do this *before* mounting if we moved the user's interactive shell into a subshell, and did a `cd` inside that subshell before exec'ing the interactive shell -- otherwise, we'd be moving the directory in which the interactive shell (which is the purpose of this program) is created to be `/`.

encfs/encfssh
rfjakob: By the way, this fails if the shell is something like `/bin/bash -x`.

encfs/encfssh
rfjakob: Where could the values come from? From `command -v` ?

encfs/encfssh
charles-dyfis-net: In addition to the arguments made in the linked articles (which I strongly suggest reading), explicit error handling is very much the Right Thing for this specific program and use case inasmuch as there's a large critical section (after the mount) following which any exit *without* an unmount is dangerously wrong.

rfjakob: My line of thinking was to do everything before the mount; But yes, probably fine to keep it like this. But probably should get some more error handling.

encfs/encfssh
charles-dyfis-net: Sure, but that's not a valid value for `$SHELL`. Quoting the POSIX spec:

> This variable shall represent a pathname of the user's preferred command language interpreter. If this interpreter does not conform to the Shell Command Language in the Shell and Utilities volume of IEEE Std 1003.1-2001, Chapter 2, Shell Command Language, utilities may behave differently from those described in IEEE Std 1003.1-2001.

Thus, it's not a command, but a path.

encfs/encfssh
charles-dyfis-net: From the user explicitly exporting `FUSE_UNMOUNT` with their own value themselves. The initial `command -v` result is only assigned if no such value exists beforehand.

charles-dyfis-net: Hmm. I was under the impression that we supported user-provided `FUSE_UMOUNT` values. That's clearly not the case, and if it's not *intended* that it be a supported interface, then this logic can be significantly simplified.

charles-dyfis-net: Reviewing the original `encfssh`, this was based on a misread; I'll pull it out.

encfs/encfssh
charles-dyfis-net: From my perspective -- as someone trying to be certain of correctness -- I'm a lot happier when I don't have to do that analysis: Expanding an array is always certain never to perform glob expansions, unwanted string-splitting, etc; thus, when proofreading code, I don't need to look back and see what values can be present in a string when expanding it unquoted to get a similar effect.

That said, if `-- -o allow_other` are the only options we'll ever need to handle, and if we can be certain that `IFS` is set at its default value, then `$options` suffices.

encfs/encfssh
```diff
@@ -28,42 +29,59 @@ if [ ! -z "$2" ]; then
...
 
 # Fall back to umount if fusermount is not available (e.g., on OS X)
-FUSE_UMOUNT="$(which 2>/dev/null fusermount)"
-FUSE_UMOUNT="${FUSE_UMOUNT:+fusermount -u}"
-FUSE_UMOUNT="${FUSE_UMOUNT:-umount}"
+fuse_umount() {
+	if command -v fusermount >/dev/null 2>&1; then
+		fusermount -u "$@"
+	else
+		umount "$@" # MacOS case
```
benrubson: FreeBSD case also, no fusermount there :)

encfs/encfssh
rfjakob: I think we'll never need anything more fancy. Please keep `$options`, I think it's good enough. Otherwise the patch looks good to me now!

charles-dyfis-net:
I've added a proposed commit as requested. That said, to propose an alternative that might make both of us happy -- what if instead of storing a string in `options`, we tracked this as a boolean, and did the following expansion?

```
allow_other=
if [ "$unenc_dir_given" = true ] && [ "$mount_public" = true ]; then
  allow_other=1
fi
encfs "$enc_dir" "$unenc_dir" ${allow_other:+ -- -o allow_other}
```

This is a win not just in that it means we don't need to suppress a shellcheck warning; it *also* improves code locality -- it's obvious on the `encfs` line exactly what the `allow_other` variable's effects can be without needing to review any context outside that line.

rfjakob:
I think the options handling is good enough as-is ;)

Merged, thanks!

PR #267, opened 2017-01-20
Improve auto unmount

benrubson:
Hello,

Here is a PR to improve auto unmount.
It is now :
- more secure because encfs will not exit if unmount failed ;
- libfuse3 compatible because it does not use `fuse_unmount_compat22` anymore, which has been removed from libfuse3.

Thank you 👍 

Ben

### DIFF
--- a/encfs/main.cpp
+++ b/encfs/main.cpp
@@ -43,11 +43,6 @@
 #include "i18n.h"
 #include "openssl.h"
 
-// Fuse version >= 26 requires another argument to fuse_unmount, which we
-// don't have.  So use the backward compatible call instead..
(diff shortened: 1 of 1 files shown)


benrubson:
Many thanks 👍

rfjakob:
Phew, this is replacing a library call with an external program spawn and adding shell injection vulnerabilities on the way. Can't we cann `fuse_unmount` somehow?

benrubson:
If you look at `fuse_unmount_compat22` code, it also calls `umount`binary :)
This is why I did it this way.
However I did not look at `fuse_unmount`, perhaps we could use it yes.

benrubson:
See #293 for a one-commit PR.
Thx !

PR #269, opened 2017-01-22
Get idle counters at once

benrubson:
Hello,

Here is a PR which get both idle counters (`usage` & `openCount`) at once.
It avoids the following race :
- get `usage`, returns 0
- _a file is opened_
- get `openCount`, returns 1
- then display _inactivity detected, but still X opened files_, whereas FS is no more inactive...

Thank you 👍 

Ben


benrubson:
Thank you for your feedback @ial0, just updated 👍

benrubson:
Many thanks 👍

benrubson:
See #294 for a one-commit PR.
Thx !

Issue #270, opened 2017-01-23
Remove timestamp from loglines ?


benrubson:
Hello,

Log lines now come with a timestamp :
```
Jan 23 11:47:51 srv1 encfs: 2017-01-23 11:47:51,372 ERROR [default] Filesystem /home/encfs/ inactivity detected, but failed to unmount
Jan 23 11:48:01 srv1 encfs: 2017-01-23 11:48:01,761 ERROR [default] Filesystem /home/encfs/ inactivity detected, but failed to unmount
Jan 23 11:48:12 srv1 encfs: 2017-01-23 11:48:12,028 ERROR [default] Filesystem /home/encfs/ inactivity detected, but failed to unmount
Jan 23 11:48:22 srv1 encfs: 2017-01-23 11:48:22,384 ERROR [default] Filesystem /home/encfs/ inactivity detected, but failed to unmount
Jan 23 11:48:32 srv1 encfs: 2017-01-23 11:48:32,740 ERROR [default] Filesystem /home/encfs/ inactivity detected, but failed to unmount
Jan 23 11:48:43 srv1 encfs: 2017-01-23 11:48:43,014 ERROR [default] Filesystem /home/encfs/ inactivity detected, but failed to unmount
Jan 23 11:48:53 srv1 encfs: 2017-01-23 11:48:53,259 ERROR [default] Filesystem /home/encfs/ inactivity detected, but failed to unmount
Jan 23 11:49:03 srv1 encfs: 2017-01-23 11:49:03,573 ERROR [default] Filesystem /home/encfs/ inactivity detected, but failed to unmount
Jan 23 11:49:14 srv1 encfs: 2017-01-23 11:49:14,025 ERROR [default] Filesystem /home/encfs/ inactivity detected, but failed to unmount
Jan 23 11:49:24 srv1 encfs: 2017-01-23 11:49:24,329 ERROR [default] Filesystem /home/encfs/ inactivity detected, but failed to unmount
```

Syslog is then unable to merge them like this :
`Jan 23 11:50:29 srv1 last message repeated 17 times`

Could we then think about removing the timestamp field from log messages ?
Does easylogging allow this ?

Many thanks 👍 

Ben

benrubson:
Solved by #271 !

benrubson:
#296 merged !

PR #271, opened 2017-01-23
Improve logging

benrubson:
Hello,

This PR solves #270.

Many thanks 👍 

Ben


benrubson:
Many thanks 👍

rfjakob:
I like getting rid of the date in daemon mode and adding colors in the foreground.
I'm not sure about the addition of "-u". Info about the user should be collected by syslog, and journald already does that pretty well. Also, ELPP_INITIALIZE_SYSLOG looks scary :)

I'd ask you to
1) Get rid of "-u"
2) Have all changes in a single commit that describes what it does (that is, mention both the removal of the date in daemon mode and the addition of colors)

benrubson:
Thank you for your feedback @rfjakob 👍 

> Info about the user should be collected by syslog, and journald already does that pretty well.

Mmmmh unfortunately [FreeBSD syslog](http://www.unix.com/man-page/FreeBSD/5/syslog.conf/) does not provide the ability to add username to generated messages.
Perhaps you were thinking about another way to do it ?

> Also, ELPP_INITIALIZE_SYSLOG looks scary :)

Why ? :) This is a documented [Easylogging feature](https://github.com/muflihun/easyloggingpp#syslog) :)

Encfs is used by many users on some of my servers, so having the username in error messages is really helpful.

benrubson:
@rfjakob perhaps instead of `-u` you would prefer `-t` to specify an alternate syslog tag ?
We would then be able to add `-t encfs/myuser`, or `-t encfs.fsdata`, or anything else we want...
As an example this feature has been implemented [into Apache](https://httpd.apache.org/docs/trunk/mod/core.html#errorlog).

rfjakob:
Yes, `-t` is much better.

It also avoids allocating a buffer whose size you don't know in advance. Can you really do this in C++ !?
```
char ident[5 + 1 + strlen(pw->pw_name) + 1];
```

benrubson:
OK @rfjakob, perfect, thank you for your feedback 👍

benrubson:
See #296 for a one-commit PR.
Thx !

Issue #272, opened 2017-01-23
EIO is returned instead of EDQUOT


benrubson:
Hello,

`EIO` is returned instead of `EDQUOT`, which can confuse the user searching for the error root cause.

Example :
```
# zfs create home/test
# zfs set userquota@root=1G home/test
# dd if=/dev/zero of=/home/test/toobig bs=128k
dd: /home/test/toobig: Disc quota exceeded
# rm /home/test/toobig 
```
```
# mkdir /home/test/one
# mkdir /home/test/two
# echo "p
password" | encfs -S /home/test/one/ /home/test/two/
# mount | grep fuse
/dev/fuse on /home/test/two (fusefs, local, synchronous)
# dd if=/dev/zero of=/home/test/two/toobig bs=128k
dd: /home/test/two/toobig: Input/output error
```
dd, writing to fuse/encfs, should have returned `EDQUOT` instead of `EIO`.

Sounds like the issue comes from this :
https://github.com/vgough/encfs/blob/v1.9.1/encfs/encfs.cpp#L625

Could we relay the underlying error please ?

Many thanks 👍 

Ben

benrubson:
Solved by #275 👍

benrubson:
Committed !

PR #275, opened 2017-01-24
Return errno in case of write failed

benrubson:
Hello,

This PR solves #272.

Strangely, I noticed that `RLOG` was modifying `errno` so that it was impossible to rely on this value right after in the code.
In my tests, `errno`, which was expected to be `EDQUOT`, was set to `EACCES` right after `RLOG` call.

**We may deeper analysis this sort of bug, as it may have side effects somewhere else in the code.**
(see #276 for this)

Thank you 👍 

Ben

### DIFF
--- a/encfs/CipherFileIO.cpp
+++ b/encfs/CipherFileIO.cpp
@@ -346,12 +346,12 @@ ssize_t CipherFileIO::readOneBlock(const IORequest &req) const {
   return readSize;
 }
 
-bool CipherFileIO::writeOneBlock(const IORequest &req) {
+int CipherFileIO::writeOneBlock(const IORequest &req) {
 
   if (haveHeader && fsConfig->reverseEncryption) {
     VLOG(1)
         << "writing to a reverse mount with per-file IVs is not implemented";
-    return false;
+    return -EPERM;
   }
 
   int bs = blockSize();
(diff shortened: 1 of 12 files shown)


ial0:
Reading `errno`, so far away from the function call you're trying to get the error from won't work.

As you mentioned RLOG may alter `errno`, any libc may alter `errno`, the logged output calls `strerror()` which may alter `errno`.  Any function that does not return an error is still allowed to alter `errno`.

The only safe way to check `errno` is directly after the call which has returned an error, in this case `::pwrite`.

As an additional note,
```
RawFileIO::truncate
RawFileIO::open
RawFileIO::getAttr
```
all return `errno`,  write is the odd one out here and should be changed for consistency alone

benrubson:
I then made `::prwite` return an `int` instead of a `bool`.
Feel free to review & comment !
Many thanks 👍

benrubson:
>write is the odd one out here and should be changed for consistency alone

Sounds like other functions do not return `errno`, such as `RawFileIO::read`.
Should we globally consistency-check all returned value ?

encfs/CipherFileIO.cpp
```diff
@@ -366,19 +366,20 @@ bool CipherFileIO::writeOneBlock(const IORequest &req) {
...
-      ok = base->write(tmpReq);
+      res = base->write(tmpReq);
     } else
-      ok = base->write(req);
+      res = base->write(req);
   } else {
     VLOG(1) << "encodeBlock failed for block " << blockNum << ", size "
             << req.dataLen;
-    ok = false;
+    res = -EIO;
```
benrubson: `EBADMSG` or `EPROTO` instead of `EIO` ?

encfs/MACFileIO.cpp
```diff
@@ -225,7 +225,7 @@ bool MACFileIO::writeOneBlock(const IORequest &req) {
   memcpy(newReq.data + headerSize, req.data, req.dataLen);
   if (randBytes > 0) {
     if (!cipher->randomize(newReq.data + macBytes, randBytes, false))
-      return false;
+      return -EIO;
```
benrubson: `EBADMSG` or `EPROTO` instead of `EIO` ?

benrubson:
@vgough @rfjakob, could we also flag this in project `1.9.2 release` please ?
Many thanks 👍

encfs/CipherFileIO.cpp
benrubson: Any thoughts about what is best here ?
Many thx !

encfs/MACFileIO.cpp
benrubson: And here ?

benrubson:
See #295 for a one-commit PR.
Thx !

PR #279, opened 2017-01-30
Support paranoia set-up in reverse mode

### LINKED ISSUE #273: Chained IV Filename in reverse mode
Hello,

I was wondering why don't we have chained IV filename allowed in reverse mode?
In its current state, two files with the same filename in different folders are encrypted to two files with the same ciphered filename. I was having a look at the source code, and I found nothing that could justify not allowing it. So I created a version where the chainedIV filename was permitted in reverse mode, tested it, and it seems to work ok.

Actually, the test was:

1. Edit FileUtils.cpp to disable the restrictions on chainedIV on reverse mode.
2. Mount encfs folder in reverse mode

> ./encfs --reverse /media/encfs/plaintext/ /media/encfs/ciphertext/
> Creating new encrypted volume.
> Please choose from one of the following options:
>  enter "x" for expert configuration mode,
>  enter "p" for pre-configured paranoia mode,
>  anything else, or an empty line will select standard mode.
> ?> x
> 
> Manual configuration mode selected.
> The following cipher algorithms are available:
> 1. AES : 16 byte block cipher
>  -- Supports key lengths of 128 to 256 bits
>  -- Supports block sizes of 64 to 4096 bytes
> 2. Blowfish : 8 byte block cipher
>  -- Supports key lengths of 128 to 256 bits
>  -- Supports block sizes of 64 to 4096 bytes
> 
> Enter the number corresponding to your choice: 1
> 
> Selected algorithm "AES"
> 
> Please select a key size in bits.  The cipher you have chosen
> supports sizes from 128 to 256 bits in increments of 64 bits.
> For example: 
> 128, 192, 256
> Selected key size: 256
> 
> Using key size of 256 bits
> 
> Select a block size in bytes.  The cipher you have chosen
> supports sizes from 64 to 4096 bytes in increments of 16.
> Or just hit enter for the default (1024 bytes)
> 
> filesystem block size: 
> 
> Using filesystem block size of 1024 bytes
> 
> The following filename encoding algorithms are available:
> 1. Block : Block encoding, hides file name size somewhat
> 2. Block32 : Block encoding with base32 output for case-insensitive systems
> 3. Null : No encryption of filenames
> 4. Stream : Stream encoding, keeps filenames as short as possible
> 
> Enter the number corresponding to your choice: 1
> 
> Selected algorithm "Block""
> 
> reverse encryption - external chained IV and MAC disabled
> Enable per-file initialization vectors?
> This adds about 8 bytes per file to the storage requirements.
> It should not affect performance except possibly with applications
> which rely on block-aligned file io for performance.
> y/[n]: y
> 
> Enable filename initialization vector chaining?
> This makes filename encoding dependent on the complete path, 
> rather then encoding each path element individually.
> [y]/n: y
> 
> 
> Configuration finished.  The filesystem to be created has
> the following properties:
> Filesystem cipher: "ssl/aes", version 3:0:2
> Filename encoding: "nameio/block", version 4:0:2
> Key Size: 256 bits
> Block Size: 1024 bytes
> Each file contains 8 byte header with unique IV data.
> Filenames encoded using IV chaining mode.
> File holes passed through to ciphertext.
> 
> Now you will need to enter a password for your filesystem.
> You will need to remember this password, as there is absolutely
> no recovery mechanism.  However, the password can be changed
> later using encfsctl.
> 
> New Encfs Password: 
> Verify Encfs Password: 
> 

3. Copy some data to plaintext folder (in my test, the linux kernel source code)
> cp -R linux-4.9.5 /media/encfs/plaintext/

4. Get a hash of plaintext files to check them later
> rhash -r -H linux-4.9.5/* > SHA1SUM

5. Get into the folder showing the ciphertext, and copy all the data to another directory out of the mount
> cd /media/encfs/ciphertext
> cp -R * /media/encfs/copy_of_ciphertext/

6. Unmount the reverse mode
> fusermount -u /media/encfs/ciphertext 

7. Copy the xml config file generated in step 1 to the copy folder of step 5, as we will be using it to mount in normal mode
> cp /media/encfs/plaintext/.encfs6.xml /media/encfs/copy_of_ciphertext/

8. Mount the copied data in normal mode
> mkdir /media/encfs/plaintext2
> encfs /media/encfs/copy_of_ciphertext /media/encfs/plaintext2

9. Check the files
> cd /media/encfs/plaintext2
> rhash -c SHA1SUM

10. The result is:
> Everything OK

So my question is: Why don't we have chained IV filename in reverse mode? It seems to work just out of the box when I disabled the controls in the source code, but since it is not allowed, I guess I'm missing something here...

Best regards !

danim7:
I propose this pull request to support paranoia set-up in reverse mode (read-only)

This closes #273 (and potentially others)
This is my first contribution to a open-source project ever....I hope I'm doing it right :)

To enable paranoia set-up in reverse mode, it is necessary to:
a. Support Filename IV chaining
b. Support External IV chaining
c. Support Block MAC headers in files

The point (a) and (b) work just out of the box when the controls avoiding using them in reverse mode are disabled.
This pull request includes support for point (c).

The tests done are:
# 
**a. Chech Filename IV chaining**
1. Mount paranoia in reverse mode: 
`encfs --reverse /media/encfs/plaintext /media/encfs/ciphertext`

2. Create two plaintext files with same filename in different folders:
```
cd /media/encfs/plaintext
mkdir d1 d2
echo aa > d1/aa
echo aa > d2/aa
```
3. Check both files are encoded with different names:
```
cd /media/encfs/ciphertext
ll  *

```
**Result**: OK, same plaintext filenames ciphered to different filenames in different folders
# 
**b. Check External IV chaining**

1. Mount paranoia in reverse mode: 
`encfs --reverse /media/encfs/plaintext /media/encfs/ciphertext`

2. Create a file:
```
cd /media/encfs/plaintext
echo aa > aa
```

3. Get checksum of ciphered file:
```
cd /media/encfs/ciphertext
sha1sum *
```

4. Rename plaintext file: 
```
cd /media/encfs/plaintext
mv aa bb
```

5. Get checksum of ciphered file:
```
cd /media/encfs/ciphertext
sha1sum *
```

6. Rename plaintext file to original filename:
```
cd /media/encfs/plaintext
mv bb aa
```

7. Get checksum of ciphered file:
```
cd /media/encfs/ciphertext
sha1sum *
```

**Result**: OK, when a plaintext file is renamed, its ciphered filename and contents change. When it is renamed back to the original filename, the ciphered filename and contents are back to the original values.
# 
**c. Check MAC headers**

1. Mount paranoia in reverse mode:
`encfs --reverse /media/encfs/plaintext /media/encfs/ciphertext`

2. Create a file in the plaintext folder:
```
cd /media/encfs/plaintext
echo aa > aa
```

3. Copy the ciphered file to a new folder:
```
cd /media/encfs/ciphertext
cp * /media/encfs/copy_of_ciphertext
```

4. Copy the config file of step 1 to the new folder:
`cp /media/encfs/plaintext/.encfs6.xml /media/encfs/copy_of_ciphertext`

5. Mount it in normal mode:
`encfs /media/encfs/copy_of_ciphertext /media/encfs/copy_of_plaintext`

6. Check sha1sum of original and copied files:
`sha1sum /media/encfs/plaintext/aa /media/encfs/copy_of_plaintext/aa`

7. Modify copied cipherfile:
`echo a >> /media/encfs/copy_of_ciphertext/*`

8. Check sha1sum of original and copied files:
`sha1sum /media/encfs/plaintext/aa /media/encfs/copy_of_plaintext/aa`

**Result**: OK, the check in step 6 matches. The check after altering the cipherfile shows an IO error
# 
**d. Non Regression Tests**

1. Mount paranoia in reverse mode:`
encfs --reverse /media/encfs/plaintext /media/encfs/ciphertext`

2. Extract and get SHA1 checksum of linux kernel sources (example with lots of different files):
```
cd /media/encfs/plaintext
tar xvf linux-4.9.5.tar.xz
rhash -r -H * > SHA1SUM
```

3. Copy ciphered files to a new folder:
```
cd /media/encfs/copy_of_ciphertext
cp -R /media/encfs/ciphertext/* .
```

4. Copy the config file of step 1 to the new folder:
`cp /media/encfs/plaintext/.encfs6.xml /media/encfs/copy_of_ciphertext`

5. Mount it in normal mode:
`encfs /media/encfs/copy_of_ciphertext /media/encfs/copy_of_plaintext`

6. Check SHA1SUM of all copied files:
```
cd /media/encfs/copy_of_plaintext
rhash -c SHA1SUM
```

**Result**: OK, SHA1SUM of all files is correct


encfs/FileUtils.cpp
```diff
@@ -1073,28 +1081,32 @@ RootPtr createV6Config(EncFS_Context *ctx,
...
+               << "\n";
+          externalIV = false;
+    }
+    selectBlockMAC(&blockMACBytes, &blockMACRandBytes, opts->requireMac, reverseEncryption);
+
+    /* Reverse mounts are read-only by default (set in main.cpp).
+     * If uniqueIV is off, writing can be allowed, because there
+     * is no header that could be overwritten */
+    if (uniqueIV == false && blockMACBytes == 0) opts->readOnly = false;
       chainedIV = selectChainedIV();
```
samrocketman: Indenting here is a little misleading because the prior conditional has the statement on the same line.

danim7: Thanks for pointing it out, fixed in commit adb464e !

dinoboy197:
It's great that you've included some tests. Would you be able to code those tests into the automated regression test suite, so they run Travis and we can be sure that your changes don't regress with future changes?

danim7:
I just added the tests to the perl script run by Travis.

encfs/MACFileIO.cpp
```diff
@@ -137,78 +141,125 @@ off_t MACFileIO::getSize() const {
...
+    IORequest tmp;
+    tmp.offset = locWithoutHeader(req.offset, blockSize(), macBytes);
+    tmp.data = mb.data + macBytes;
+    tmp.dataLen = req.dataLen - macBytes;
+
+    // get the data from the base FileIO layer
+    ssize_t readSize = base->read(tmp);
+    if (readSize == 0) {
+      MemoryPool::release(mb);
+      return 0;
```
benrubson: Code below will fail if `readSize < 0`.
It could occur in case of storage IO error.
In this case you should return `readSize` (as done above).
Thx !

danim7: Done, thx !

benrubson: Perfect thx 👍

encfs/MACFileIO.cpp
```diff
@@ -137,78 +141,125 @@ off_t MACFileIO::getSize() const {
...
+
+    MemoryPool::release(mb);
+
+    return readSize;
+
+  } else {	  //reverse = true
+
+    if (randBytes > 0) {
+      throw Error(_("MAC rand bytes not yet implemented in reverse mode"));
+      return -1;
```
benrubson: I think you should `RLOG()` (instead of `throw Error`), and return `-EPERM`.

encfs/MACFileIO.cpp
```diff
@@ -137,78 +141,125 @@ off_t MACFileIO::getSize() const {
...
+    return readSize + macBytes;
 
-  return readSize;
+  }
 }
 
 bool MACFileIO::writeOneBlock(const IORequest &req) {
+  if (reverse) {
+    throw Error(_("MAC write not implemented in reverse mode"));
+    return false;
```
benrubson: I think you should `RLOG()` (instead of `throw Error`).

encfs/MACFileIO.cpp
```diff
@@ -248,6 +299,11 @@ bool MACFileIO::writeOneBlock(const IORequest &req) {
 }
 
 int MACFileIO::truncate(off_t size) {
+  if (reverse) {
+    throw Error(_("MAC write not implemented in reverse mode"));
+    return -1;
```
benrubson: I think you should `RLOG()` (instead of `throw Error`), and return `-EPERM`.

danim7: All three done, thanks for reviewing !

danim7:
Merged with current 'master' as of 2017-08-06 and conflicts resolved. Any chance to get this pulled?

encfs/MACFileIO.cpp
```diff
@@ -255,6 +305,11 @@ bool MACFileIO::writeOneBlock(const IORequest &req) {
 }
 
 int MACFileIO::truncate(off_t size) {
+  if (reverse) {
+    RLOG(ERROR) << "MAC write not implemented in reverse mode";
+    return -EPERM;
```
benrubson: Not sure how to handle truncate. Any thoughts ?

encfs/MACFileIO.cpp
```diff
@@ -137,84 +141,130 @@ off_t MACFileIO::getSize() const {
...
+      return readSize + macBytes;
 
-  return readSize;
+    }
 }
 
 bool MACFileIO::writeOneBlock(const IORequest &req) {
+  if (reverse) {
+    RLOG(ERROR) << "MAC write not implemented in reverse mode";
+    return false;
```
benrubson: As now we have `--reversewrite`,  perhaps we would have to do a MAC comparison before writing the block, as in `readOneBlock()` above ?

benrubson:
@danim7 do you think you would be able to handle the few changes above ?
Many thanks 👍

danim7:
Hello @benrubson

It has been a while since I worked with encfs, and I may be wrong about
this, but I don't think it is a easy change.

You will have to get a whole block to be able to compute the MAC
successfully. That means you will need some kind of buffering before
writing a block, to treat the cases where an application or the operating
system gives you a partial block.

Another particular case would be if an application writes a partial block
on purpose, expecting you will write it to disk. In that case, you will
have to read the whole block from disk, modify it partially, recheck the
MAC and write it if successful.

I don't know how you may choose one or the other write strategy at the
encfs level...there are quite a few corner cases.

Unfortunately I have almost no free time right now... Please feel free to
modify the pull request (maintainers shall be allowed to do it) or pull
from my repository ;)

Thanks !

benrubson:
Thank you @danim7 for your answer.

Yes partial blocks must be handled carefully.

However, 2 thoughts regarding this :
- there is a good chance that an application will provide the whole block to write, as as soon as a part of the block is modified, MAC is also modified ; that said, application could also provide MAC in one write operation, and the other part of the block in another operation (I think for example about rsync delta-syncing a ciphered-repo to a reverse EncFS, to update the plain version) ;
- I think we already handle such cases in the non-reverse version where partial blocks must first be read, so that their MAC get correctly updated.

chancemon:
I would be thrilled if these changes could be finalized and merged.

benrubson:
_This would need to be re-worked completely, closing for now..._

Issue #280, opened 2017-01-31
git version breaks years old CLI API


mhogomchungu:
Discussion for this problem is here: https://github.com/tomm/cryptkeeper/issues/23

rfjakob:
Yep, looks like https://github.com/vgough/encfs/commit/c3a7da5eff4055e77dc9404b0c15945485232bf2 should be reverted.

PR #282, opened 2017-02-05
Revert "-S" ABI change

rfjakob:
Revert c3a7da5eff4055e77dc9404b0c15945485232bf2 and enforce empty password ban.

Context: https://github.com/tomm/cryptkeeper/issues/23 https://github.com/vgough/encfs/issues/280


rfjakob:
@vgough I will merge this tomorrow unless you object

jrchamp:
Does it make sense to consider requiring the password length to be greater than 1? Or is that just a pointless, compatibility-breaking change?

rfjakob:
@jrchamp I'm already afraid that enforcing non-empty passwords will cause some complaints, I'd rather not take this further ;)

Issue #283, opened 2017-02-10
Extended attributes on symlinks cannot be read on decrypted directories


dinoboy197:
Using the latest release as of the filing of this issue (1.9.1), If a symlink to a regular file exists in a decrypted directory, the extended attributes on the target of the link cannot be read.

Steps to reproduce:

```
# verify version
encfs --version

# remove old runs
rm -rf /tmp/source
fusermount -u /tmp/decrypted
fusermount -u /tmp/encrypted

# setup - create a new directory with a symlink to the /bin/tar utility
mkdir -p /tmp/source /tmp/encrypted /tmp/decrypted
ls -al /bin/tar
ln -s /bin/tar /tmp/source

# encrypt /tmp/source into /tmp/encrypted
encfs --standard --reverse -o ro /tmp/source /tmp/encrypted

# decrypt /tmp/encrypted into /tmp/decrypted
ENCFS6_CONFIG=/tmp/source/.encfs6.xml encfs --standard /tmp/encrypted /tmp/decrypted

echo "Testing:"
echo "First, try rsync from decrypted directory to another location without xattrs."
rsync -avHA /tmp/decrypted/ /tmp/copy1/
echo "Works properly; exit code is $?"

echo "Now, try rsync from decrypted directory to another location with xattrs."
rsync -avHAX /tmp/decrypted/ /tmp/copy2/
echo "Fails; exit code is $?"

echo "And, just to be sure that copying xattrs from original source works ok..."
rsync -avHAX /tmp/source/ /tmp/copy3/
echo "Works properly; exit code is $?"
```

ekalin:
That seems to be the same problem as my report #247 . When a symlink pointing to a non existent file (there is not /bin/tar in the encrypted filesystem), one gets an error when trying to get the extended attributes of the symlink (because encfs uses a system call to get the attributes of the target file, not of the symlink itself).

Issue #247, opened 2016-11-12
Extended attributes problem with dangling symlinks


ekalin:
On a normal filesystem, this is what happens when one tries to list attributes in a dangling symlink:
```
$ ln -s inexistent_file.txt bad_link

$ attr -l bad_link
[No output]

$ attr -l inexistent_file.txt
attr_list: No such file or directory
Could not list "(null)" for inexistent_file.txt
```

But on an encfs mount, there is an error:
```
$ ln -s inexistent_file.txt bad_link

$ attr -l bad_link
attr_list: No such file or directory
Could not list "(null)" for bad_link

$ attr -l inexistent_file.txt
attr_list: No such file or directory
Could not list "(null)" for inexistent_file.txt
```

It seems that instead of calling llistxattr() (which lists xattrs from the link itself), listxattr() (which lists xattrs from the pointed-to file) is called instead. This is an strace from the "attr -l bad_link" command in a normal filesystem:

`llistxattr("bad_link", "", 65536)       = 0`

And here's the strace of the encfs process of the same command in an encfs mount:

`listxattr("/tmp/e/cryptstore/tTj5KY7uJx3Uf-PDeR0bIV7b", 0x1f7df80, 65536) = -1 ENOENT (No such file or directory)`

dinoboy197:
I believe I'm seeing a manifestation of the same problem in #283 . @ekalin , can you reproduce my problem in #283?

PR #285, opened 2017-02-12
Turn on all integration tests in Travis build

dinoboy197:
I'm hoping to file a bug fix in the next several days for #247 and #283, and so I thought it would be a good idea to get not only [the CPP code tests](https://github.com/vgough/encfs/blob/master/encfs/test.cpp) executing in travis CI builds, but also all of [the Perl integration tests](https://github.com/vgough/encfs/tree/master/tests) as well. This should help to prevent regressions with future code changes. For instance, in my fix for #247 and #283 I'll likely add a new regression test into that suite.

I did have to turn `sudo` mode on to do this, so that `fuse` can be initialized. However, the remainder of the build was untouched and still runs fine. It also still runs in under 90 seconds :)

Also there are two newlines at the end of the travis config file; I removed one of them so now there is only one.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: cpp
 
-sudo: false
+sudo: true
 
 compiler:
   - clang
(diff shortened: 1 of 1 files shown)


rfjakob:
Awesome!

dinoboy197:
@rfjakob thanks for merging so quickly!

PR #289, opened 2017-02-17
Change Spanish initials to English in es_ES.po

antcc:
Change 's' to 'y' when asked to create a directory if it does not already exist.
It made no sense to keep Spanish initials, since the program has the English ones hardcoded. Previously the user would get an error message with no information about the problem, until they figured out that they should have used English initials.


rfjakob:
Merged, thanks!

PR #293, opened 2017-03-09
Improve auto unmount

benrubson:
Hello,

Here is a PR to improve auto unmount.

It does not use `fuse_unmount_compat22` anymore, which :
- has been removed from libfuse3 ;
- has some issues under FreeBSD.

Thank you 👍 

Ben

### DIFF
--- a/encfs/main.cpp
+++ b/encfs/main.cpp
@@ -43,11 +43,6 @@
 #include "i18n.h"
 #include "openssl.h"
 
-// Fuse version >= 26 requires another argument to fuse_unmount, which we
-// don't have.  So use the backward compatible call instead..
-extern "C" void fuse_unmount_compat22(const char *mountpoint);
-#define fuse_unmount fuse_unmount_compat22
-
 /* Arbitrary identifiers for long options that do
  * not have a short version */
 #define LONG_OPT_ANNOTATE 513
@@ -715,6 +710,11 @@ static void *idleMonitor(void *_arg) {
   const int timeoutCycles = 60 * arg->idleTimeout / ActivityCheckInterval;
   int idleCycles = -1;
 
+  bool unmountres = false;
+
+  // We will notify when FS will be unmounted, so notify that it has just been mounted
+  RLOG(INFO) << "Filesystem mounted: " << arg->opts->mountPoint;
+
   pthread_mutex_lock(&ctx->wakeupMutex);
 
   while (ctx->running) {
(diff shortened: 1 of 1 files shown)


benrubson:
@rfjakob could we also take a look at this when U'll have time ?
Many thx 👍

benrubson:
@rfjakob could you also have a look at this when you'll have time please ?
Thank you very much for your help 👍

encfs/main.cpp
```diff
@@ -768,9 +772,18 @@ static bool unmountFS(EncFS_Context *ctx) {
...
     // Time to unmount!
-    RLOG(WARNING) << "Unmounting filesystem due to inactivity: "
-                  << arg->opts->mountPoint;
-    fuse_unmount(arg->opts->mountPoint.c_str());
-    return true;
+    int rc=1;
+    if (access("/bin/umount", F_OK) != -1)
+      rc = system(("/bin/umount "+std::string(arg->opts->mountPoint)).c_str());
+    else
+      rc = system(("/sbin/umount "+std::string(arg->opts->mountPoint)).c_str());
```
samrocketman: I wonder if it's worth it to search `$PATH` if mount is not found in the "normal" locations.

Also, how would this port to Windows, Mac OS, or FreeBSD?

benrubson: Thank you for your review Sam !

The risk of searching through `$PATH` is to run a malicious `umount`, this is why I used only hard-coded paths.

FreeBSD :
```
# which umount
/sbin/umount
```

Mac OS :
```
# which umount
/sbin/umount
```

Do you have some systems in mind with `umount` elsewhere ?

Regarding Windows, good question.
Seems that encfs4win [calls](https://github.com/jetwhiz/encfs4win/blob/v1.10.1-RC11/encfs/main.cpp#L957) `fuse_unmount` but [disables](https://github.com/jetwhiz/encfs4win/blob/v1.10.1-RC11/encfs/main.cpp#L51) `fuse_unmount_compat22`.

benrubson: Thank you for your review Sam !

The risk of searching through `$PATH` is to run a malicious `umount`, this is why I used only hard-coded paths.

FreeBSD :
```
# which umount
/sbin/umount
```

Mac OS :
```
# which umount
/sbin/umount
```

Do you have some systems in mind with `umount` elsewhere ?

Regarding Windows, good question.
Seems that encfs4win [calls](https://github.com/jetwhiz/encfs4win/blob/v1.10.1-RC11/encfs/main.cpp#L957) `fuse_unmount` but [disables](https://github.com/jetwhiz/encfs4win/blob/v1.10.1-RC11/encfs/main.cpp#L51) `fuse_unmount_compat22`, as this `fuse_unmount` call goes to [Dokany](https://github.com/dokan-dev/dokany/tree/v1.0.3) library.

benrubson: > - we can't call `fuse_unmount` as we don't have a fuse handle (`fuse_main` does not return it) ;

Well, according to encfs4win call to `fuse_unmount` (second parameter to `NULL`), and according to [libfuse source code](https://github.com/libfuse/libfuse/blob/fuse_2_9_5/lib/helper.c#L250), we could clearly set this second parameter to `NULL`.

Let me do some tests, however I think I will still face the drawbacks listed in the first post.

samrocketman: I notice encfs4win is a complete fork.  Is encfs not designed to run on Windows as well?  I guess not since it requires a whole fork.

benrubson:
So here is a fuse-only method, which is then Windows compatible.

samrocketman:
Cool :-) looks good to me.  I'll let someone with more seasoned eyes than mine do the final approval and merge.

benrubson:
Thank you very much Sam 👍

benrubson:
Hi @rfjakob !
What about merging this PR ?
Many thx 👍

rfjakob:
Let's go.

benrubson:
Thank you !

PR #294, opened 2017-03-09
Get idle counters at once

benrubson:
Hello,

Here is a PR which get both idle counters (`usage` & `openCount`) at once.
It avoids the following race :
- get `usage`, returns 0
- _a file is opened_
- get `openCount`, returns 1
- then display _inactivity detected, but still X opened files_, whereas FS is no more inactive...

Thank you 👍 

Ben

### DIFF
--- a/encfs/main.cpp
+++ b/encfs/main.cpp
@@ -718,15 +718,19 @@ static void *idleMonitor(void *_arg) {
   pthread_mutex_lock(&ctx->wakeupMutex);
 
   while (ctx->running) {
-    int usage = ctx->getAndResetUsageCounter();
+    int usage, openCount;
+    ctx->getAndResetUsageCounter(&usage, &openCount);
 
     if (usage == 0 && ctx->isMounted())
       ++idleCycles;
-    else
+    else {
+      if (idleCycles >= timeoutCycles)
+        RLOG(INFO) << "Filesystem no longer inactive: "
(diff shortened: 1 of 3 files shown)


encfs/main.cpp
```diff
@@ -718,15 +718,19 @@ static void *idleMonitor(void *_arg) {
...
+    int usage, openCount;
+    ctx->getAndResetUsageCounter(&usage, &openCount);
 
     if (usage == 0 && ctx->isMounted())
       ++idleCycles;
-    else
+    else {
+      if (idleCycles >= timeoutCycles)
+        RLOG(INFO) << "Filesystem no longer inactive: "
+                   << arg->opts->mountPoint;
```
rfjakob: Why is this message wrapped in the if?

benrubson: Let's imagine the FS was inactive `(idleCycles >= timeoutCycles)`, but unmount failed for some reason.
FS is then still mounted.
Activity to the FS comes back, we then log a "Filesystem no longer inactive" message (so only if it has been inactive before :)

benrubson: PR #293 will have logged the "Filesystem inactive but unmount failed" message.

rfjakob: Ok.

benrubson:
Thx 👍

Issue #281, opened 2017-02-01
(stream|block)(Encode|Decode) never return false ?


benrubson:
Hello,

Going through the code, I noticed that following function never return false :
```
streamEncode
streamDecode
blockEncode
blockDecode
```

Whereas calling functions (`writeOneBlock` / `blockWrite`...) test their returned value.

Is it something to correct ?

Many thanks 👍 

Ben

benrubson:
Solves by #297 !

benrubson:
Committed !

Issue #276, opened 2017-01-24
errno is sometimes not correctly used / not used at all


benrubson:
Hello,

Sometimes `errno` is not correctly used, its value may be modified by other functions before it is used.
For example here : [RawFileIO.cpp#L154](https://github.com/vgough/encfs/blob/v1.9.1/encfs/RawFileIO.cpp#L154)...

And sometimes, `errno` is not used at all in returned values where it could be interesting to have it.
For example here : [RawFileIO.cpp#L206](https://github.com/vgough/encfs/blob/master/encfs/RawFileIO.cpp#L206), here : [RawFileIO.cpp#L189](https://github.com/vgough/encfs/blob/master/encfs/RawFileIO.cpp#L189)...

So here is a bug report not to forget to deeper-analysis this.

Thank you 👍 

Ben

benrubson:
See #295.

benrubson:
Committed !

PR #295, opened 2017-03-09
Code sanity check

benrubson:
Hello,

This PR tries to improve code sanity.
I was surprised some functions are called but their returned values are never checked, which could lead to inconsistencies / corrupted data.

This PR then tries to help :
It solves #272, making `RawFileIO::write` return an `int` instead of a `bool`.
It also makes `readOneBlock` and `initHeader` return `-EBADMSG` when decode fails, as in above.
It  solves #281, making (stream|block)(Encode|Decode) return false when needed.
It verifies open/read/write/truncate returned values.
It corrects some wrong `errno` usage (#276).

Of course feel free to comment / make a deep sanity check on your side, of course I could have missed some points to correct...

Thank you 👍 

Ben

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,7 +20,7 @@ before_script:
   - ./ci/setup.sh
 
 script:
-  - ./ci/check.sh
+  - ./ci/build.sh
 
 addons:
   apt:
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -10,7 +10,11 @@ Compiling EncFS
 
 EncFS uses the CMake toolchain to create makefiles.
 
-Steps to build EncFS:
+Quickest way to build and test EncFS :
+
+    ./build.sh
+
+Or following are the detailed steps to build EncFS:
 
     mkdir build
     cd build
@@ -29,11 +33,11 @@ encrypted filesystem and run tests on it:
     make integration
 
 The compilation process creates two executables, encfs and encfsctl in
-the encfs directory.  You can install to in a system directory via
+the encfs directory.  You can install to in a system directory via:
 
     make install
 
-. If the default path (`/usr/local`) is not where you want things
+If the default path (`/usr/local`) is not where you want things
 installed, then set the CMAKE_INSTALL_PREFIX option when running cmake.  Eg:
 
     cmake .. -DCMAKE_INSTALL_PREFIX=/opt/local
--- a/build.sh
+++ b/build.sh
@@ -12,4 +12,9 @@ then
 fi
 
 make -j2 -C build
+make test -C build
+make integration -C build
+
+echo
+echo 'Everything looks good, you can install via "make install -C build".'
 
--- a/ci/build.sh
+++ b/ci/build.sh
@@ -16,6 +16,7 @@ cd build
 cmake .. ${CFG}
 make -j2
 make test
+make integration
 
 cd ..
 


benrubson:
_still some work pending on this, please wait before merging :)_

benrubson:
>Of course feel free to comment / make a deep sanity check on your side, of course I could have missed some points to correct...

Would be nice if some of the contributors could review this PR.
We would then be able to think about merging it.
Many thx 👍 

(_in production here since more than 2 months with no issue_)

benrubson:
Anyone ? :)

benrubson:
To easily review this PR, you should take each of the 8 commits one by one, as each one works on a specific task / has its own goal.
Of course all of them are related to code sanity check / better error handling 👍 
Thank you !

encfs/BlockFileIO.cpp
```diff
@@ -171,11 +171,11 @@ ssize_t BlockFileIO::read(const IORequest &req) const {
   return result;
 }
 
-bool BlockFileIO::write(const IORequest &req) {
+int BlockFileIO::write(const IORequest &req) {
   CHECK(_blockSize != 0);
 
   off_t fileSize = getSize();
-  if (fileSize < 0) return false;
+  if (fileSize < 0) return fileSize;
```
samrocketman: This changes the boolean check.  Normally, it would return false if `fileSize < 0` so line 256 would break.  However you now evaluate the negative `int` in the conditional so whether it returns a positivie or negative it will be evaulated as `true` and execute break.

Is this intended to change the behavior?

benrubson: Thank you for your review @samrocketman 👍 
So with this change, if BlockFileIO::write return value `<= 0`, it means it failed.
If return value `== 0`, it means it succeeded.
This function does not return value `> 0`.
So line 256 will break only if write failed, and the error code will be properly returned 👍 
Generally in C, return code `0` means it succeeded :)

benrubson:
@rfjakob I would really like if you could review this PR when U'll have some time.

To make things easier, each of the 8 commits should be analyzed one by one, as each one works on a specific task / has its own goal.
All of them are related to code sanity check / better error handling 👍 

Thank you !

encfs/RawFileIO.cpp
```diff
@@ -269,6 +270,14 @@ int RawFileIO::truncate(off_t size) {
     knownSize = true;
   }
 
+  if (fd >= 0 && canWrite) {
```
vgough: why move this block?

CMakeLists.txt
```diff
@@ -93,6 +93,12 @@ check_cxx_source_compiles ("#include <sys/types.h>
   " XATTR_ADD_OPT)
 set (CMAKE_REQUIRED_QUIET ${CMAKE_REQUIRED_QUIET_SAVE})
 
+# Check if we have fdatasync.
+include (CheckCXXSourceCompiles)
```
vgough: CHECK_FUNCTION_EXISTS (fdatasync HAVE_FDATASYNC) ?

encfs/RawFileIO.cpp
benrubson: Because `fdatasync` (or `fsync`) could modify `errno` set by `truncate`, and what we want in `errno` here is the return code of `truncate`, not `*sync`.

CMakeLists.txt
benrubson: Of course if it works, let's go with this nicer solution 👍 
Was not aware of it (well, did not look after it :)

encfs/BlockFileIO.cpp
```diff
@@ -147,6 +147,10 @@ ssize_t BlockFileIO::read(const IORequest &req) const {
       }
 
       ssize_t readSize = cacheReadOneBlock(blockReq);
+      if (readSize < 0) {
+        result = readSize;
```
vgough: can you document in the method comment what return value means now?
positive (or zero?) value == success,
negative value (or zero?) == failure?

encfs/BlockFileIO.cpp
```diff
@@ -191,7 +195,9 @@ bool BlockFileIO::write(const IORequest &req) {
   if (req.offset > fileSize) {
     // extend file first to fill hole with 0's..
     const bool forceWrite = false;
-    padFile(fileSize, req.offset, forceWrite);
+    int res = padFile(fileSize, req.offset, forceWrite);
+    if (res)
```
vgough: so a positive value is a failure here?  Is a positive value from ::write considered a failure?  Seems like a change in logic here.

encfs/BlockFileIO.cpp
benrubson: Just committed :
_Returns the number of bytes read, or -errno in case of failure._

encfs/BlockFileIO.cpp
```diff
@@ -251,8 +260,8 @@ bool BlockFileIO::write(const IORequest &req) {
     }
 
     // Finally, write the damn thing!
-    if (!cacheWriteOneBlock(blockReq)) {
-      ok = false;
+    res = cacheWriteOneBlock(blockReq);
+    if (res) {
```
vgough: res < 0 ?  why exit the loop on a positive block result?

encfs/BlockFileIO.cpp
benrubson: `padFile` returns 0 in case of success, <0 (`-errno`) in case of error.
What it does is simply returning the return code of `cacheWriteOneBlock`.

vgough:
I made it through the first quarter.  I'm a little concerned with the int error logic -- it would be helpful if there was a consistent definition of <0 means error, >=0 success (or something along those lines).  But the code seems to sometimes treat !=0 as failure, sometimes as success, etc.  I think there's a lot of good work in here, it does worry me that the error value handling needs closer review.

encfs/BlockFileIO.cpp
vgough: Explicitly checking "res < 0" would help, since otherwise I have to know which functions return positive values or never greater than 0..

encfs/RawFileIO.cpp
vgough: ahh, side effects.  Thanks.

encfs/BlockFileIO.cpp
benrubson: OK I'll then clarify return codes in a new commit.

benrubson: Commit done 👍

encfs/BlockFileIO.cpp
benrubson: I just committed a comment to help : write methods return  the very first result returned by `RawFileIO::write` so :
 0 in case of success, or -errno in case of failure.

benrubson:
So read methods return <0 (`-errno`) in case of failure, or number of bytes read in case of success.
Write methods return <0 (`-errno`) in case of failure, or 0 (no matter the number of written bytes, we already know how many bytes we wanted to write).

benrubson:
To easily understand the changes, I think you should review one commit by one commit.
As each one has its own goal.
It would be much easier to read :)

benrubson:
OK I just made return codes easier to read 👍

samrocketman:
@vgough in general only ==0 is success and !=0 is failure.  This because intuitively there's typically only one reason for success: success; but there could be many reasons for failure.

vgough:
If the commits are separable, then using separate PRs makes it easier on the reviewer, since it's easier to track status in github's review system.  I'll continue reviewing here.

benrubson:
These commits are not really separable, as it's a whole sanity check review.
They are however more understandable if they are taken one by one :)
But anyway, thank you for reviewing @vgough !

benrubson:
Lots of new commits in master, should I try to rebase this against master @vgough, or do you still plan to temporarily revert the other commits ?
Thank U 👍

vgough:
I haven't worked through this entirely, and some fixes have now already gone in (eg move fdatasync after error capture).  So I'm afraid you'll have to rebase, or we can pick or a piece at a time.

benrubson:
OK, I'll then rebase it now.

benrubson:
OK, just rebased 👍

encfs/BlockFileIO.cpp
```diff
@@ -338,35 +357,38 @@ void BlockFileIO::padFile(off_t oldSize, off_t newSize, bool forceWrite) {
...
       req.offset = newLastBlock * _blockSize;
       req.dataLen = newBlockSize;
       memset(mb.data, 0, req.dataLen);
-      cacheWriteOneBlock(req);
+      res = cacheWriteOneBlock(req);
     }
   }
 
   if (mb.data != nullptr) {
     MemoryPool::release(mb);
```
vgough: memory pool management seems to keep getting in the way of simpler cleanup on failure.  I'll rework the memorypool API to improve this.

encfs/BlockFileIO.cpp
```diff
@@ -401,20 +425,19 @@ int BlockFileIO::truncateBase(off_t size, FileIO *base) {
     req.data = mb.data;
 
     ssize_t rdSz = cacheReadOneBlock(req);
+    if (rdSz < 0) {
+      res = rdSz;
+    }
 
     // do the truncate
```
vgough: please move comment inside of else.  
} else if (...) {
  //
}

encfs/CipherFileIO.cpp
```diff
@@ -98,20 +98,23 @@ bool CipherFileIO::setIV(uint64_t iv) {
   } else if (haveHeader) {
     // we have an old IV, and now a new IV, so we need to update the fileIV
     // on disk.
+    // ensure the file is open for read/write..
```
vgough: why move this block?

encfs/CipherFileIO.cpp
```diff
@@ -447,19 +465,24 @@ int CipherFileIO::truncate(off_t size) {
...
+      }
     }
 
     // can't let BlockFileIO call base->truncate(), since it would be using
     // the wrong size..
     res = BlockFileIO::truncateBase(size, nullptr);
 
-    if (res == 0) {
-      base->truncate(size + HEADER_SIZE);
+    if (!(res < 0)) {
```
vgough: wouldn't "res >= 0" be clearer?

encfs/MACFileIO.cpp
```diff
@@ -260,8 +260,8 @@ int MACFileIO::truncate(off_t size) {
 
   int res = BlockFileIO::truncateBase(size, nullptr);
 
-  if (res == 0) {
-    base->truncate(locWithHeader(size, bs, headerSize));
+  if (!(res < 0)) {
```
vgough: res >= 0

vgough: or,
if (res < 0) { return res; }
return base->truncate(...);

encfs/RawFileIO.cpp
```diff
@@ -220,14 +221,17 @@ bool RawFileIO::write(const IORequest &req) {
   ssize_t bytes = req.dataLen;
   off_t offset = req.offset;
 
+  int eno = 0;
   while ((bytes != 0) && retrys > 0) {
+    errno = 0;
```
vgough: instead of setting errno, can we eno into the "writeSize < 0" branch?  pwrite will return -1 if it sets errno.

encfs/encfs.cpp
```diff
@@ -712,10 +708,11 @@ int encfs_fsync(const char *path, int dataSync, struct fuse_file_info *file) {
 }
 
 int _do_write(FileNode *fnode, unsigned char *ptr, size_t size, off_t offset) {
-  if (fnode->write(offset, ptr, size)) {
-    return size;
+  int res = fnode->write(offset, ptr, size);
+  if (res < 0) {
+    return res;
   }
-  return -EIO;
```
vgough: would be nice if write returned the size on success.

vgough:
Looks good overall.  I have only a couple minor questions.

encfs/RawFileIO.cpp
benrubson: No as we need `eno` on line 247 : if we exit loop because we did not write all data (`retrys == 0)`, it is nice to check `errno`, we never know.

benrubson:
OK, I answered / solved your question @vgough, as well as some `int` / `(s)size_t` corrections.
I hope all this PR will make the code "stronger".

encfs/RawFileIO.cpp
vgough: I don't follow.. pwrite will return -1 if it sets errno.
So the only time you need to worry about errno is in the path "if (writeSize < 0)" which returns unconditionally.  So how can eno ever be != at line 247?

I'd rather not mutate global variables if not necessary.
You could rewrite as:  
eno = 0;
... pwrite
if (writeSize < 0) {
  eno = errno;
...
}

But then it's pretty clear that eno is only needed for that path.

benrubson: I just wondered why `write` would not have succeeded after having written something 10 times...
So I thought, as it did not manage to write what we want, perhaps it wrote only a small piece of our data and gently set `errno` to explain why.
Of course, not sure however it can happen...

Do you have a situation in mind where `write` could succeed 10 times without writing all the data ?

benrubson:
I did a huge mistake and erased all my work...
What a mess !
I have it as a `.patch` file (without the last modifications), I will then rework it entirely...
I can't see how to retrieve my work, if ever you have an idea...
Perhaps you have my PR synced somewhere @vgough ?

encfs/RawFileIO.cpp
benrubson: _Feel free to answer me here @vgough, even if #380 has been opened to "replace" this PR._

Issue #244, opened 2016-11-03
Possible thread deadlock in eassylogging


ial0:
While looking into issue #214 I ran clang's thread sanitizer,  this found a potential deadlock within easylogging++.  I'm not sure if this has any bearing on #214 or not could it could be part of the issue.

```
==================
WARNING: ThreadSanitizer: lock-order-inversion (potential deadlock) (pid=3933)
  Cycle in lock order graph: M11 (0x7d480000be90) => M13 (0x7d4400009d90) => M17 (0x7d580000fc10) => M11

  Mutex M13 acquired here while holding mutex M11 in main thread:
    #0 pthread_mutex_lock <null> (encfs+0x0000004e1285)
    #1 el::base::threading::internal::Mutex::lock() /home/ial/Devel/Encfs/encfs/internal/easylogging++.h:996:45 (encfs+0x0000005680ec)
    #2 el::base::threading::internal::ScopedLock<el::base::threading::internal::Mutex>::ScopedLock(el::base::threading::internal::Mutex&) /home/ial/Devel/Encfs/encfs/internal/easylogging++.h:1031:54 (encfs+0x000000567f09)
    #3 el::base::RegisteredLoggers::get(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, bool) /home/ial/Devel/Encfs/encfs/internal/easylogging++.h:3673:45 (encfs+0x00000056b26c)
    #4 el::Loggers::getLogger(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, bool) /home/ial/Devel/Encfs/encfs/internal/easylogging++.h:5825:47 (encfs+0x0000005e01e5)
    #5 el::Loggers::reconfigureLogger(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, el::Configurations const&) /home/ial/Devel/Encfs/encfs/internal/easylogging++.h:5846:47 (encfs+0x0000005e0069)
    #6 encfs::initLogging(bool) /home/ial/Devel/Encfs/encfs/encfs/Error.cpp:20:3 (encfs+0x0000005df77e)
    #7 main /home/ial/Devel/Encfs/encfs/encfs/main.cpp:546:3 (encfs+0x00000055a7fe)

    Hint: use TSAN_OPTIONS=second_deadlock_stack=1 to get more informative warning message

  Mutex M17 acquired here while holding mutex M13 in main thread:
[...]

Issue #214, opened 2016-09-15
Segfault in encfs::FileNode::open(int)


m13253:
Description:
After upgrading to encfs 1.9-2, using pam_encfs to automatically decrypt home directory causes coredump.
I can not confirm whether it is a bug of encfs or a bug of pam_encfs.
Log attached in [log.txt](https://bugs.archlinux.org/task/50789?getfile=14505).
I reproduced it 3 times. But sorry I did not manage to debug the program, because I can not afford any data loss.
Downgrading to 1.8.1-7 solves the problem.

Additional info:
ArchLinux x86_64
encfs: 1.9-2
pam_encfs: 0.1.4.4-4

Cross post: https://bugs.archlinux.org/task/50789

vgough:
I haven't been able to reproduce this issue.  Makes me wonder if it depends on a particular version or FUSE or kernel.  Committed an additional null check in this path just in case.

m13253:
Thank you.
I will try to reproduct this problem later this week.

Could you please tell me what additional info do I need to help diagnoze this problem? (In addition to a core dump)

vgough:
It would be helpful if you can reproduce this without pam and document the steps you took to see the issue.

Also, please provide the encfs configuration used.

m13253:
> Also, please provide the encfs configuration used.

`encfs -S --idle=1 -v /home/.Private/brilliant /home/brilliant -- -o allow_other,allow_other,nonempty`

> It would be helpful if you can reproduce this without pam and document the steps you took to see the issue.

I will try to catch a core dump next time, both with pam and without pam. (Between Oct 1-7, on which I will have holidays)
The crash happens when KDE is loading. It will not crash when I log into it with Virtual Terminal.
Although I have no core dump, there is a log, which is provided in the link.

I am terribly sorry for having you been waiting so long.

m13253:
I reinstalled the old version, and produced a core dump.
I will analyze it. What extra action would you like me to do with my core dump?

m13253:
This is a transcript of the disassembly:
```asm
; encfs::FileNode::open(int) const ()
push rbp
push rbx
mov  rbx,rdi
mov  ebp esi
call pthread_mutex_lock@plt
mov  rdi,QWORD PTR [rbx+0x38]
mov  esi,ebp
mov  rax,QWORD PTR [rdi] ; <== crash here, rdi=0x00000000
call QWORD PTR [rax+0x38]
mov  rdi,rbx
mov  ebp,eax
[...]

PR #296, opened 2017-03-09
Improve logging

benrubson:
Hello,

This PR does 2 things :
- it removes timestamp from log lines when syslog in use (solves #270) ;
- it adds `-t` option to tune syslog tag.

Thank you 👍 

Ben

### DIFF
--- a/encfs/main.cpp
+++ b/encfs/main.cpp
@@ -491,11 +491,6 @@ void *encfs_init(fuse_conn_info *conn) {
   // set fuse connection options
   conn->async_read = true;
 
-  if (ctx->args->isDaemon) {
-    // Switch to using syslog.
-    encfs::rlogAction = el::base::DispatchAction::SysLog;
-  }
-
   // if an idle timeout is specified, then setup a thread to monitor the
   // filesystem.
   if (ctx->args->idleTimeout > 0) {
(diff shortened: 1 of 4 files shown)


benrubson:
@rfjakob could we think about merging this easy PR please ? Many thx 👍

encfs/main.cpp
```diff
@@ -190,6 +191,7 @@ static bool processArgs(int argc, char *argv[],
   out->isVerbose = false;
   out->idleTimeout = 0;
   out->fuseArgc = 0;
+  out->syslogTag.assign("encfs");
```
rfjakob: Maybe a stupid question, but why not simply

    out->syslogTag = "encfs";

encfs/main.cpp
```diff
@@ -265,6 +269,9 @@ static bool processArgs(int argc, char *argv[],
       case 'S':
         out->opts->useStdin = true;
         break;
+      case 't':
+        out->syslogTag.assign(optarg);
```
rfjakob: Why not use "="?

encfs/main.cpp
benrubson: Because option `p)` some lines below uses `assign` :)
I updated with `=` which is more convenient 👍

dickerpulli:
When is 1.9.2 going to be released?

benrubson:
@rfjakob / @vgough will decide, but would be nice to try to solve #214 / #244 I think.
And merge #293 / #295 / #309 / #317.
Here's my wishlist for 1.9.2 :)

PR #297, opened 2017-03-09
Make (stream|block)(Encode|Decode) return false

benrubson:
Hello,

This PR solves issue #281, making (stream|block)(Encode|Decode) return false when needed.

Thank you 👍 

Ben


benrubson:
See #295.

Issue #242, opened 2016-10-29
Read write for reverse mount


syaifulnizamyahya:
I use reverse mount as below

```
echo <password> | ENCFS6_CONFIG='/boot/acd_cli/config/.encfs6.xml' encfs -S --reverse -o rw -o allow_other -o uid=99 -o gid=100 /mnt/user/Amazon/local/ /mnt/user/Amazon/.local/
```

Unfortunately, .local folder is read only. How do I mount .local as read write? I want to move .local contents to a remote folder(in the cloud). I set my encryption to be file independent. Meaning, encrypted files can be placed anywhere and yet can still be decrypted.

EldonMcGuinness:
I wish this was possible as in my use case it makes it a bit cumbersome.

diskmaster23:
Did anyone ever answer this question? I am having the same issue.

pcjonathan:
I too wish this was possible.

rfjakob:
There seems to be a bug where the very first mount is read-only. Subsequent mounts are read-write.

```
$ encfs --version
	encfs version 1.9.1
$ mkdir a b
$ touch a/foo
$ encfs --reverse $PWD/a $PWD/b
	Creating new encrypted volume.
	Please choose from one of the following options:
	 enter "x" for expert configuration mode,
	 enter "p" for pre-configured paranoia mode,
	 anything else, or an empty line will select standard mode.
	?> 

	Standard configuration selected.

	Configuration finished.  The filesystem to be created has
	the following properties:
	Filesystem cipher: "ssl/aes", version 3:0:2
	Filename encoding: "nameio/block", version 4:0:2
	Key Size: 192 bits
	Block Size: 1024 bytes
	File holes passed through to ciphertext.

	Now you will need to enter a password for your filesystem.
	You will need to remember this password, as there is absolutely
	no recovery mechanism.  However, the password can be changed
	later using encfsctl.

	New Encfs Password: 
	Verify Encfs Password: 
$ ls -l b
	total 4
	-rw-rw-r--. 1 jakob jakob    0  2. Jan 13:02 B9ZT7iQhSf0LupxqVCsS2Edg
	-rw-rw-r--. 1 jakob jakob 1264  2. Jan 13:03 ii328JX2cjDCDeNjVQqKGAU1
$ rm b/B9ZT7iQhSf0LupxqVCsS2Edg 
	rm: cannot remove 'b/B9ZT7iQhSf0LupxqVCsS2Edg': Read-only file system
[...]

PR #299, opened 2017-03-09
Allow read/write in standard reverse mode

benrubson:
Hello,

This PR solves #242, allowing read/write mode in standard reverse mode.

Thx !

Ben


benrubson:
As you can see this PR initially broke the reverse `writesDenied()` test, which is expected because conf is created in reverse mode, so `uniqueIV` are disabled, so writes are allowed.
We would need a specific reverse conf with (`uniqueIV == true`) to test that reverse writes are disabled.
So I disabled this test for the moment, of course feel free to improve the test 👍

benrubson:
See #300 for a one-commit PR.
Thx !

PR #301, opened 2017-03-09
Allow read/write in standard reverse mode

benrubson:
Hello,

This PR solves #242, allowing read/write mode in standard reverse mode.

Of course this PR brakes the reverse writesDenied() test, which is expected because conf is created in reverse mode, so uniqueIV are disabled, so writes are allowed.
We would need a specific reverse conf with (uniqueIV == true) to test that reverse writes are disabled.
So I disabled this test for the moment, of course feel free to improve it.

Thx 👍 

Ben


rfjakob:
Looks good, thanks.

Issue #291, opened 2017-03-04
 enfcs 1.9.1 on SUSE Leap 42.2 undefined references to `el::base::elStorage'


Aikhjarto:
Building enfcs 1.9.1 on SUSE Leap 42.2 fails with undefined reference:
```
[  112s] CMakeFiles/encfs.dir/encfs/base64.cpp.o: In function `el::base::MessageBuilder::operator<<(char const*)':
[  112s] /home/abuild/rpmbuild/BUILD/encfs-1.9.1/internal/easylogging++.h:4637: undefined reference to `el::base::elStorage'
[  112s] CMakeFiles/encfs.dir/encfs/base64.cpp.o: In function `el::base::MessageBuilder::operator<<(unsigned int)':
[  112s] /home/abuild/rpmbuild/BUILD/encfs-1.9.1/internal/easylogging++.h:4631: undefined reference to `el::base::elStorage'
[  112s] CMakeFiles/encfs.dir/encfs/base64.cpp.o: In function `el::base::LogFormat::updateFormatSpec()':
[  112s] /home/abuild/rpmbuild/BUILD/encfs-1.9.1/internal/easylogging++.h:2355: undefined reference to `el::base::utils::s_currentHost'
[  112s] /home/abuild/rpmbuild/BUILD/encfs-1.9.1/internal/easylogging++.h:2349: undefined reference to `el::base::utils::s_currentUser'
[  112s] CMakeFiles/encfs.dir/encfs/base64.cpp.o: In function `el::base::MessageBuilder::operator<<(char const*)':
[  112s] /home/abuild/rpmbuild/BUILD/encfs-1.9.1/internal/easylogging++.h:4637: undefined reference to `el::base::elStorage'
[  112s] CMakeFiles/encfs.dir/encfs/base64.cpp.o: In function `el::base::Writer::initializeLogger(std::string const&, bool, bool)':
[  112s] /home/abuild/rpmbuild/BUILD/encfs-1.9.1/internal/easylogging++.h:5037: undefined reference to `el::base::elStorage'
[  112s] /home/abuild/rpmbuild/BUILD/encfs-1.9.1/internal/easylogging++.h:5020: undefined reference to `el::base::elStorage'
[  112s] /home/abuild/rpmbuild/BUILD/encfs-1.9.1/internal/easylogging++.h:5023: undefined reference to `el::base::elStorage'
[...]

PR #308, opened 2017-03-10
Inhibit --no-undefined added by some compilers

benrubson:
Hi,

This PR solves #291.

Thank you 👍 

Ben

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -40,6 +40,10 @@ else ()
   endif()
 endif ()
 
+# Some compilers add -Wl,--no-undefined which leads to easylogging undefined references (see #291)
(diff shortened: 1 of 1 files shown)


benrubson:
This PR is a workaround, I did not found a way to make encfs compile with `CMAKE_MODULE_LINKER_FLAGS` having `-Wl,--no-undefined`.
I found [this](https://forum.qt.io/topic/42546/solved-issue-with-easylogging), perhaps it may help, but I did not manage to go further with this.
Thx 👍

rfjakob:
No, I don't think this is a good idea.

benrubson:
Let's close this as it is a bad idea.

Issue #303, opened 2017-03-10
Minimum required cmake version?


Aikhjarto:
`CMakeLists.txt` says `cmake_minimum_required(VERSION 2.8)`

`ci/install-cmake.sh` pulls cmake-3.1.3

I'm having trouble building with cmake 2.8 and wonder whats the real required minimum version of cmake for building encfs.

benrubson:
According to different closed issues (for example [this one](https://github.com/vgough/encfs/issues/226)), you should a newer cmake version, or try to install the `gettext` package.

Aikhjarto:
Ok, with 3.1 it works well.
So maybe in `CMakeLists.txt` the minimum required version should be raised.

benrubson:
I think we need at least 2.8.8 for gettext functions.
But not so sure...
@rfjakob should confirm :)

benrubson:
I gave 2.8.9 a try on a Debian 7.11 server (with `gettext` installed), it fails :
```
# cmake --version
cmake version 2.8.9
# cmake ..
** Assuming that GNU CXX uses -std=c++11 flag for C++11 compatibility.
CMake Error at /usr/share/cmake-2.8/Modules/FindOpenSSL.cmake:263 (list):
  list GET given empty list
Call Stack (most recent call first):
  CMakeLists.txt:64 (find_package)


CMake Error at /usr/share/cmake-2.8/Modules/FindOpenSSL.cmake:264 (list):
  list GET given empty list
Call Stack (most recent call first):
  CMakeLists.txt:64 (find_package)


CMake Error at /usr/share/cmake-2.8/Modules/FindOpenSSL.cmake:266 (list):
  list GET given empty list
Call Stack (most recent call first):
  CMakeLists.txt:64 (find_package)


CMake Error at /usr/share/cmake-2.8/Modules/FindOpenSSL.cmake:268 (list):
  list GET given empty list
Call Stack (most recent call first):
  CMakeLists.txt:64 (find_package)


-- Using local TinyXML2 copy
-- Enabled syslog logging support
CMake Warning at CMakeLists.txt:131 (find_package):
  By not providing "FindIntl.cmake" in CMAKE_MODULE_PATH this project has
  asked CMake to find a package configuration file provided by "Intl", but
  CMake did not find one.

  Could not find a package configuration file provided by "Intl" with any of
  the following names:

    IntlConfig.cmake
    intl-config.cmake

[...]

PR #309, opened 2017-03-11
Update cmake

benrubson:
Hello,

Here is a PR which update cmake prerequisites, see #303.

Thx 👍 

Ben


benrubson:
@rfjakob and this one ? :) Thx 👍

benrubson:
@rfjakob, what about merging this PR please ?
Many thanks 👍

rfjakob:
Problem here is that you cannot build on debian stable anymore. I don't know, but there should be a better solution

benrubson:
Strange, do you have a debian stable build log ?

rfjakob:
No, but Debian stable has cmake 3.0.2 ( https://packages.debian.org/jessie/cmake ). I think the goal should be to make it "just work" there.

benrubson:
It will build with cmake 3.0.2, cmake 3.2.0 is just an "advice" to run configuration flawlessly (it correctly detects Intl etc...).
In addition, next stable Debian (which should come shortly) has cmake 3.7.
And you can backport cmake 3.6 on Jessie 👍

rfjakob:
Why doesn't it run flawlessly on cmake 3.0.2? Is the encfs build process so special?

benrubson:
cmake 3.0.2 does not detect Intl, leading to :
```
CMake Warning at CMakeLists.txt:131 (find_package):
  Could not find module FindIntl.cmake or a configuration file for package
  Intl.
  Adjust CMAKE_MODULE_PATH to find FindIntl.cmake or set Intl_DIR to the
  directory containing a CMake configuration file for Intl.  The file will
  have one of the following names:
    IntlConfig.cmake
    intl-config.cmake
```

Whereas cmake >= 3.2.0 detects it, which is better, no error output.

rfjakob:
And if we ship the file FindIntl.cmake with encfs?

benrubson:
Seems as a bad workaround to me, as we would have to maintain this "external" file :|
It will still build with cmake 3.0.2, you will simply get an advice to run a newer version :
```
+if (${CMAKE_MAJOR_VERSION}.${CMAKE_MINOR_VERSION} LESS 3.2)
+  message(WARNING "You should use cmake 3.2.0 or newer for configuration to run correctly.")
+endif()
```

rfjakob:
Looks pretty compact: https://github.com/Kitware/CMake/blob/master/Modules/FindIntl.cmake
Maybe we can pull it into the cmake config file?

rfjakob:
No, probably better to keep it a separate file.

I mean, we ship our own copies of tinyxml and easylogging to make it easy to build encfs. We should continue to try making it easy.

benrubson:
According to previous releases, Debian Stretch should be there in June or July.
https://wiki.debian.org/DebianStretch
Then, what about moving to 3.2.0 (with 3.0.2 (and lower !) still usable) ?

rfjakob:
Haha, no, encfs is 10 years old, we shouldn't need a not-yet-released distro to compile it without errors

benrubson:
Should be OK now @rfjakob 👍 
It correctly runs with cmake 3.0.2, and this is now the minimum version advised (in cmake configuration file and INSTALL.md).
However, I give Travis cmake 3.1.0, as it is the first x86_64 compiled version.

benrubson:
What about this @rfjakob ?
Many thx 👍

benrubson:
Should be even better now @rfjakob 👍 
It correctly runs with cmake 3.0.2, and this is now the minimum version advised (in cmake configuration file and INSTALL.md).
I also give Travis cmake version 3.0.2. As this x86_64 compiled version is not available online, I compiled it and pushed it to the repository.
Thank you 👍

rfjakob:
Thanks ben, i'm currently travelling abroad, i'll be back next week and
will try to merge this

benrubson:
@rfjakob, just to remind you this PR :)
Many thx 👍

benrubson:
@rfjakob do you think we could merge this PR ?
Thank you very much for your help 👍

benrubson:
Many thanks @rfjakob 👍

benrubson:
Mmmmh sounds like CircleCI is failing, I did not notice we used both Travis and CircleCI.
Do you know why we use both of them ?
Perhaps we could remove CircleCI ?

rfjakob:
I'm not sure why we have both, probably because Travis could not run FUSE tests at some point?

But I'd rather have the failure fixed than removing the tests. It's an additional platform that is tested, after all.

benrubson:
I think we should then `make check` instead of `make test` :
https://github.com/vgough/encfs/blob/master/circle.yml#L13

benrubson:
#327 should at least help us.

benrubson:
OK we now have some more output regarding failing tests in CircleCI, perfect.
CircleCI does not seem to be correctly configured, at least as Travis is (no `modprobe fuse` etc...).
Certainly this is why it is failing.

Issue #19, opened 2014-09-12
switch to libgcrypt


vgough:
Current dev tree supports either OpenSSL, Botan, or CommonCrypto.

Having worked now with these libraries, and tested a few more, I'm standardizing on libgcrypt for encfs2.

benrubson:
@danim7 has proposed a switch to libgcrypt in #310.
One simple question, what's the benefit of using libgcrypt instead of openssl ?
Thx 👍

PR #310, opened 2017-03-12
Switch to libgcrypt

danim7:
Hello,

I was checking what is necessary to have a secure 2.0 version that fixes the issues from the audit. One of the issues flagged as 2.x candidate is switching to libgcrypt library (Issue #19 ). It could be interesting to do this step before implementing a change in the data format or in the the cipher mode of operation. Therefore, I started to play with the gcrypt library and tried to "translate" the SSL_Cipher.cpp file from openSSL to libgcrpyt, building upon the gcrypt branch.

This pull request introduces support to libgcrypt library. It passes the tests from "make test" and ./test.sh perl tests. Additionally, it can successfully cipher a whole filesystem to be deciphered by the openssl version, and vice versa.

Some points about this version:
- In my machine, libgcrypt seems to be as fast as openssl in encrypting / decrypting.
- In my machine, libgcrypt seems to be quite slower than openssl in generating random strong numbers. Actually, it is necessary to tell the library if we want strong or "nonce" random numbers to keep a reasonable performance.
- Libgcrypt doesn't support blowfish with a key size different than 128 bits: https://gnupg.org/documentation/manuals/gcrypt/Available-ciphers.html#Available-ciphers
- In this version, the openssl support has been deactivated in "Cipher.cpp" file, it can be activated back by uncommenting the REF_MODULE section. Ideally, there shall be a mechanism to choose between different cipher implementations.
- Other than that, the library is quite simple to use and it has many available symmetric ciphers.

I hope this is useful. Please, play around with it and tell me what you think.

This pull request is against the gcrypt branch.

### DIFF
--- a/encfs/GCrypt_Cipher.cpp
+++ b/encfs/GCrypt_Cipher.cpp
@@ -1,0 +1,921 @@
+/*****************************************************************************
+ * Author:   Valient Gough <vgough@pobox.com>
+ *           danim7 (https://github.com/danim7)
+ *****************************************************************************
+ * Copyright (c) 2004, Valient Gough
+ *
+ * This program is free software: you can redistribute it and/or modify it
+ * under the terms of the GNU Lesser General Public License as published by the
+ * Free Software Foundation, either version 3 of the License, or (at your
+ * option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE.  See the GNU Lesser General Public License
+ * for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#include "internal/easylogging++.h"
+#include <cstring>
+#include <pthread.h>
+#include <string>
+#include <sys/mman.h>
+#include <sys/time.h>
+
+#include "Cipher.h"
+#include "Error.h"
+#include "Interface.h"
+#include "Mutex.h"
+#include "Range.h"
+#include "intl/gettext.h"
+#include "GCrypt_Cipher.h"
+
+using namespace std;
+
+namespace encfs {
+
+const int MAX_KEYLENGTH = 32;  // in bytes (256 bit)
+const int MAX_IVLENGTH = 16;   // 128 bit (AES block size, Blowfish has 64)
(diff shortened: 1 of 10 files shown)


encfs/GCrypt_Cipher.cpp
```diff
@@ -0,0 +1,917 @@
...
+  for (int i = size - 1; i; --i) buf[i] ^= buf[i - 1];
+}
+
+
+
+/** Partial blocks are encoded with a stream cipher.  We make multiple passes on
+ the data to ensure that the ends of the data depend on each other.
+*/
+bool GCrypt_Cipher::streamEncode(unsigned char *buf, int size, uint64_t iv64,
+                              const CipherKey &ckey) const {
```
benrubson: This function (and others) never return false (some kind related to #295).
Perhaps you could return false when they fail ?
And correctly check their returned values when calling them ?
Thank you 👍

danim7: Done, I implemented the modifications you did in #295 to the SSL_Cipher.cpp file to my GCrypt_Cipher.cpp

vgough:
Thanks for this work, I've been wanting to support grcypt as the backend after having tried a number of backends in another branch.  I'd like to make it a build-time flag to select which to use.

encfs/GCrypt_Cipher.cpp
benrubson: Thx, perfect 👍

LinuxDonald:
Any timeline for merging this?

madscientist42:
You flatly (And I **_DO_** mean this...) don't want to do this unless it's an option somehow to choose either.

People keep thinking in terms of their ideas on what's good/cool in crypto but never ONCE think in terms of what can/can't be used for FIPS-140-2 and higher levels of certification for any of this.

OpenSSL is certified with a given Cryptographic "Module" and devices can be ultimately certified with that in there so long as your cryptographic uses only rely on the approved module's functions.

libgcrypt does **_NOT_** support anything of the sort.  So...  PLEASE do not do this without thinking LONG and hard on what you're doing here.

madscientist42:
You want either OpenSSL or WolfSSL at this time as they are the only ones in the space right now with approvals.  You can support MORE, but you need to support one or the other with things.  Right now, encfs is the only one that can be certified for FIPS compliance at this time.  All the others use unapproved symmetric key, asymmetric key (which is not currently approved for anything...), or a hashing algo like scrypt which isn't approved.

benrubson:
_This would need to be re-worked completely, closing for now..._

PR #312, opened 2017-03-13
Re-list fuse operations

benrubson:
Hello,

This PR correctly re-orders fuse operations according to the documentation : http://libfuse.github.io/doxygen/structfuse__operations.html

In addition, it identifies those removed from fuse 3.0.0.

Thx 👍 

Ben

### DIFF
--- a/encfs/main.cpp
+++ b/encfs/main.cpp
@@ -555,8 +555,8 @@ int main(int argc, char *argv[]) {
   memset(&encfs_oper, 0, sizeof(fuse_operations));
 
   encfs_oper.getattr = encfs_getattr;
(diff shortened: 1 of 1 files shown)


rfjakob:
Hmm, I don't think there is a "correct" order of filling the struct.

benrubson:
I just put them according to the order given in the documentation :)
http://libfuse.github.io/doxygen/structfuse__operations.html
I went through all operations and was easier to read having them sorted, so I simply pushed the modification :)

benrubson:
And this one @rfjakob ? ^^

rfjakob:
I don't think the case is strong enough for this one. Merging this will make diffing old and new encfs code more difficult.

benrubson:
OK, understood :)
Thx for your feedback anyway 👍

Issue #315, opened 2017-03-20
FreeBSD does not pass tests


benrubson:
Hi,

```
# uname -sr
FreeBSD 11.0-RELEASE-p7
```

```
# ./build/encfs --version
encfs version 1.9.1
```

```
# ./test.sh 
+ '[' '!' -d build ']'
+ perl -MTest::Harness -e '$$Test::Harness::verbose=1; runtests @ARGV;' tests/normal.t.pl tests/reverse.t.pl
tests/normal.t.pl ... 54/104 
#   Failed test 'corrupted file with MAC returns read error: '
#   at tests/normal.t.pl line 80.
tests/normal.t.pl ... 58/104 # Looks like you failed 1 test of 104.
tests/normal.t.pl ... Dubious, test returned 1 (wstat 256, 0x100)
Failed 1/104 subtests 
	(less 2 skipped subtests: 101 okay)
tests/reverse.t.pl .. 3/26 
#   Failed test 'ciphertext and decrypted size of file grown to 10 bytes'
#   at tests/reverse.t.pl line 131.
tests/reverse.t.pl .. 9/26 
#   Failed test 'decrypted files are identical'
#   at tests/reverse.t.pl line 72.
tests/reverse.t.pl .. 22/26 # Looks like you failed 2 tests of 26.
tests/reverse.t.pl .. Dubious, test returned 2 (wstat 512, 0x200)
Failed 2/26 subtests 

Test Summary Report
-------------------
tests/normal.t.pl (Wstat: 256 Tests: 104 Failed: 1)
  Failed test:  56
  Non-zero exit status: 1
tests/reverse.t.pl (Wstat: 512 Tests: 26 Failed: 2)
  Failed tests:  8, 12
  Non-zero exit status: 2
Files=2, Tests=130, 16 wallclock secs ( 0.04 usr  0.00 sys + 14.97 cusr  0.45 csys = 15.45 CPU)
Result: FAIL
Failed 2/2 test programs. 3/130 subtests failed.
```

Of course underlying FS is ZFS.
But results are the same using a tmpfs.

@enodata perhaps you noticed this maintaining FreeBSD _encfs_ port ?

Thank you 👍 

Ben

benrubson:
Normal test is now OK with #317.

Reverse still fails, I'm pretty sure this is because Fuse in FreeBSD does not support `entry_timeout=0` and `attr_timeout=0` (set by `--nocache` option).
[...]

PR #317, opened 2017-03-21
Correct test with FreeBSD

benrubson:
Hello,

This PR partially solves the reverse test with FreeBSD.

Thx 👍 

Ben

### DIFF
--- a/tests/reverse.t.pl
+++ b/tests/reverse.t.pl
@@ -132,6 +132,11 @@ sub grow {
         last unless $ok;
     }
     ok($ok, "ciphertext and decrypted size of file grown to $i bytes");
+    close($pfh);
+    close($vfh);
+    close($cfh);
+    close($dfh);
+    unlink("$plain/grow"); 
 }
 
 sub largeRead {


benrubson:
@rfjakob what about merging this one too please ? Thank U !

tests/normal.t.pl
```diff
@@ -28,10 +28,10 @@ sub runTests
     my $hardlinks = 1;
     if($mode eq 'standard')
     {
-        &mount("--standard");
+        &mount("--standard -o direct_io");
     } elsif($mode eq 'paranoia')
     {
-        &mount("--paranoia");
+        &mount("--paranoia -o direct_io");
```
rfjakob: I don't think this is a good idea. The tests should run with the options that are actually used in the wild.

tests/reverse.t.pl
```diff
@@ -132,6 +132,11 @@ sub grow {
         last unless $ok;
     }
     ok($ok, "ciphertext and decrypted size of file grown to $i bytes");
+    close($pfh);
+    close($vfh);
+    close($cfh);
+    close($dfh);
+    unlink("$plain/grow");
```
rfjakob: This is ok.

tests/normal.t.pl
benrubson: Well, when you look at the test which fails with FreeBSD (58/104, see #315), the fact that it fails is quite expected :
- the _fuse_ file is created, FreeBSD puts it into its cache, then it is closed ;
- the underlying file is corrupted putting some garbage into it ;
- the _fuse_ file is reopened, and fully read from the cache.
Of course the cache has not been updated when underlying file has been modified, as I'm not sure it can work this way.

I think this behaviour is quite "expected". Don't you think so ?
Should we update the test so that it will fail without `direct_io` ?

rfjakob: How do network file systems on FreeBSD handle this, sshfs for example? There, you cannot assume that the file does not change behind your back.

rfjakob: (This works on Linux because, by default, Linux flushes the file from cache when it is opened. Check out kernel_cache in `man mount.fuse`.)

benrubson: Strangely enough `MOUNT_FUSEFS(8)` has the same option :
```
     Besides the above mount options, there is a set of pseudo-mount options
     which are supported by the Fuse library. One can list these by passing -h
     to a Fuse daemon. Most of these options have effect only on the behaviour
     of the daemon (that is, their scope is limited to userspace). However,
     there are some which do require in-kernel support.  Currently the options
     supported by the kernel are:

     direct_io
             Bypass the buffer cache system

     kernel_cache
             By default cached buffers of a given file are flushed at each
             open(2).  This option disables this behaviour
```

rfjakob: Interesting. Does not seem to be working, though.

benrubson: From https://github.com/freebsd/freebsd/blob/stable/11/sys/fs/fuse/fuse_kernel.h#L117 :
`FOPEN_KEEP_CACHE: don't invalidate the data cache on open`

However `FOPEN_KEEP_CACHE` does not seem to be implemented at all :
https://github.com/freebsd/freebsd/blob/stable/11/sys/fs/fuse/fuse_node.c#L270

Certainly there is something like the following missing :
```
if (!(fuse_open_flags & FOPEN_KEEP_CACHE)) {
	fuse_io_invalbuf(vp, td);
}

benrubson: So I was OK, I patched FreeBSD fuse adding the 3 previous lines of code, and it works.
I'll then correct this PR accordingly and open a FreeBSD issue.

benrubson:
@rfjakob you can now merge this :)
Thank you 👍

benrubson:
Many thx 👍

Issue #253, opened 2016-12-16
Unable to read encfs config via named pipe


antofthy:
For a VERY long time..  (Since 2009)  I have used a special system where I do not store the ".encfs5" with the encrypted data, but store it in a special (secure) database along with the keys to the encrypted data.
To pass the configuration file to encfs I create a named pipe which I put into the "ENCFS6_CONFIG" environment variable.   I then start the "encfs", which reads the config file from the named pipe as I write to it.   This way the config file is NEVER written to the disk in the clear.

For example -- (shell pseudo code - real code is written in perl)
   mkpipe temporary_named_pipe
   cat "the real config file info" >>  temporary_named_pipe &
   ENCFS6_CONFIG="temporary_named_pipe"  encfs  --stdin encfs_dir decoded_dir  <<<"$password"
   rm temporary_named_pipe

The problem is suddenly I get errors from encfs when it reads its config from a named pipe...

2016-12-16 20:59:43,610 ERROR [default] Failed to load config file /tmp/ks_lij9g9
2016-12-16 20:59:43,610 ERROR [default] Found config file /tmp/ks_lij9g9, but failed to load - exiting

If I replace the named pipe with a plain text file with the exact same config, but no other changes, it all works perfectly fine.   If interested the code is available in  "ks"  http://www.ict.griffith.edu.au/anthony/software/#ks

Why did it change?

 Why can't encfs read from a named pipe?

What can I do about it?

rfjakob:
Running under strace (`ENCFS6_CONFIG="$PWD/pipe" strace encfs $PWD/a $PWD/b`) reveals this:

```
open("/tmp/tmp.9pa6Ln8X3Y/pipe", O_RDONLY) = 3
fstat(3, {st_mode=S_IFIFO|0664, st_size=0, ...}) = 0
lseek(3, 0, SEEK_SET)                   = -1 ESPIPE (Illegal seek)
read(3, "<?xml version=\"1.0\" encoding=\"UT"..., 4096) = 1264
fstat(3, {st_mode=S_IFIFO|0664, st_size=0, ...}) = 0
lseek(3, 0, SEEK_END)                   = -1 ESPIPE (Illegal seek)
lseek(3, 0, SEEK_CUR)                   = -1 ESPIPE (Illegal seek)
lseek(3, 0, SEEK_SET)                   = -1 ESPIPE (Illegal seek)
close(3)
```

[...]

PR #319, opened 2017-04-11
Read config into memory and pass its content to tinyxml2 (#253)

johbo:
Trying to address issue #253.

From investigating the comments I found that it seems to be best to read the configuration file into memory first and then pass the content to tinyxml2, using `XmlDocument.Parse()` instead of `XmlDocument.LoadFile()`.

### DIFF
--- a/encfs/XmlReader.cpp
+++ b/encfs/XmlReader.cpp
@@ -23,6 +23,8 @@
 #include <algorithm>  // for remove_if
 #include <cstring>    // for NULL
 #include <memory>     // for shared_ptr
+#include <fstream>    // for ifstream
+#include <sstream>    // for ostringstream
 
 #include <tinyxml2.h>  // for XMLElement, XMLNode, XMLDocument (ptr only)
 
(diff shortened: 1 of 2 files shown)


johbo:
Also tried to add a test  for this problem, but it seems that I did not yet get it quite right: https://github.com/johbo/encfs/commit/59b571a5e33195bbd85ebf18c57a96233e980d70

I did observe this output which seems to be caused by the added test:

```
  Parse errors: Tests out of sequence.  Found (49) but expected (50)
                Tests out of sequence.  Found (50) but expected (51)
                Tests out of sequence.  Found (51) but expected (52)
                Tests out of sequence.  Found (52) but expected (53)
                Tests out of sequence.  Found (53) but expected (54)
Displayed the first 5 of 60 TAP syntax errors.
```

benrubson:
You also have to increase counter at line 5 of `tests/normal.t.pl` by the number of tests you add.

johbo:
Thanks for the hint. I've also noticed that I did get into some extra trouble because one of the calls to `ok` was running inside of the child process. Seems this did confuse the test runner. After changing these bits it did at least pass on my debian box.

rfjakob:
Looks good to me.

PR #320, opened 2017-04-11
[ignore, travis testing]

rfjakob:


samrocketman:
Why not include cmake version as a part of the regular build?

PR #327, opened 2017-05-28
Make CircleCI run make check

benrubson:
As for Travis, make CircleCI run make check.


benrubson:
OK we now have some more output regarding failing tests in CircleCI, perfect.

Issue #329, opened 2017-06-07
Mac OS Sierra `make check` failure


hughperkins:
```
2/2 Test #2: scriptedtests ....................***Failed   20.32 sec
/Users/me/mygit/encfs/tests/normal.t.pl ... ok
Can't exec "attr": No such file or directory at /Users/me/mygit/encfs/tests/reverse.t.pl line 87.

#   Failed test 'symlink to '/' extended attributes can be read (return code was -1)'
#   at /Users/me/mygit/encfs/tests/reverse.t.pl line 89.
#          got: '-1'
#     expected: '0'
Can't exec "attr": No such file or directory at /Users/me/mygit/encfs/tests/reverse.t.pl line 87.

#   Failed test 'symlink to 'foo' extended attributes can be read (return code was -1)'
#   at /Users/me/mygit/encfs/tests/reverse.t.pl line 89.
#          got: '-1'
#     expected: '0'
Can't exec "attr": No such file or directory at /Users/me/mygit/encfs/tests/reverse.t.pl line 87.

#   Failed test 'symlink to '/1/2/3/4/5/6/7/8/9/10/11/12/13/14/15/15/17/18' extended attributes can be read (return code was -1)'
#   at /Users/me/mygit/encfs/tests/reverse.t.pl line 89.
#          got: '-1'
#     expected: '0'
Can't exec "attr": No such file or directory at /Users/me/mygit/encfs/tests/reverse.t.pl line 87.

#   Failed test 'symlink to '!§$%&/()\<>\#+=' extended attributes can be read (return code was -1)'
#   at /Users/me/mygit/encfs/tests/reverse.t.pl line 89.
#          got: '-1'
#     expected: '0'
Can't exec "attr": No such file or directory at /Users/me/mygit/encfs/tests/reverse.t.pl line 87.

#   Failed test 'symlink to '/var/folders/fn/asdasd/T//encfs-reverse-tests-bRQS/plain/foo' extended attributes can be read (return code was -1)'
#   at /Users/me/mygit/encfs/tests/reverse.t.pl line 89.
#          got: '-1'
#     expected: '0'
# Looks like you failed 5 tests of 25.
/Users/me/mygit/encfs/tests/reverse.t.pl .. 
Dubious, test returned 5 (wstat 1280, 0x500)
Failed 5/25 subtests 

```

hughperkins:
this is on latest github `master`:
```
$ git log -n 3
[...]

PR #330, opened 2017-06-08
Correctly detect attr binary during tests

benrubson:
Hello,

This issue should solve issue #329 as Mac OS uses `xattr` command instead of `attr`.

Ben

### DIFF
--- a/encfs/main.cpp
+++ b/encfs/main.cpp
@@ -252,7 +252,7 @@ static bool processArgs(int argc, char *argv[],
     // 'o' : arguments meant for fuse
     // 't' : syslog tag
     int res =
-        getopt_long(argc, argv, "HsSfvdmi:o:t:", long_options, &option_index);
+        getopt_long(argc, argv, "HsSfvdmi:o:t:V", long_options, &option_index);
 
     if (res == -1) break;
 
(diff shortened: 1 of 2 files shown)


benrubson:
@rfjakob I think you can safely merge this and close #329 :)
Thank you again 👍

benrubson:
More generic version which now also supports FreeeBSD.
FreeBSD uses `lsextattr` binary.
However, it also uses different includes to support xattr, which are not yet implemented by encfs.

benrubson:
@rfjakob let's merge this one and close #329 ?
Thank you (again) 👍 
Ben

benrubson:
@rfjakob what about this one (and then close #329) ?
Thank you again for your support 👍 
Ben

rfjakob:
Hmm the commit message does not match what the patch does. What is the "V" command line argument?

benrubson:
It shows the encfs version, to which I added compilation options (`HAVE_XATTR`).
This flag is then used by reverse test to know if encfs has `HAVE_XATTR` or not.

rfjakob:
I have commited the part that detects the attr binary as 8d8506f2eb1d7aa0cec1c4fb6eff11273b950b9b .

I'm not sure about the "detect if we were built without xattr support" part, but in any case, this would be a separate commit.

benrubson:
Thank you very much @rfjakob.
I'll then submit another PR for the builds without xattr support.
(as for example FreeBSD builds are made without xattr support)

PR #332, opened 2017-06-19
Define _DIRENT_HAVE_D_TYPE in one place

coypoop:
for the operating systems that do not define it.
This is a functional change as _DIRENT_HAVE_D_TYPE had the OS
checks in one place but not another.

I've only tested basic usage and only on NetBSD.

### DIFF
--- a/encfs/DirNode.cpp
+++ b/encfs/DirNode.cpp
@@ -42,6 +42,10 @@
 #include "Error.h"
 #include "Mutex.h"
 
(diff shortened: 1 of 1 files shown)


vgough:
Thanks for identifying a problem and proposing a fix!  I'd like to move these platform capability checks into CMake, so that we can condition on detected features rather than platforms.  I've taken the liberty of doing that in #372.

PR #338, opened 2017-06-25
Detect binaries without xattr support during tests

benrubson:
Hello,

This PR follows #330, it corrects reverse test so that it now detects if EncFS has been compiled with xattr support.
It then avoids to fail xattr tests if they are not supported.
This is the case with FreeBSD which uses different includes to support xattr, not yet implemented by encfs.

Thank you 👍 

Ben

### DIFF
--- a/encfs/main.cpp
+++ b/encfs/main.cpp
@@ -245,7 +245,7 @@ static bool processArgs(int argc, char *argv[],
     // 'o' : arguments meant for fuse
     // 't' : syslog tag
     int res =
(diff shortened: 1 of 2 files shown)


benrubson:
Hi @rfjakob, what about this fix ?
Thank you very much 👍

benrubson:
I'll slightly modify this one once #342 merged.

benrubson:
_Directly Included in #342, easier to manage._

Issue #341, opened 2017-06-28
Copying symlinks on OSX results in error message


erikman:
I have some issues using `encfs` on osx. I am using latest trunk version of encfs and OSXFuse 3.6.0 on OSX 10.12.5

Steps to reproduce:
```
mkdir raw
mkdir enc
encfs -v -f raw enc -- -d
Standard mode, password 1234
```

From finder copy a symbolic link into `enc`.

Expected result:
The symbolic link is copied.

Observed result:
 The symbolic link is copied but I also get an error message in a popup box:

```
The operation can’t be completed because one or more required items can’t be found.
(Error code -43)
```

Log file from when I do the copy operation for a link called `_mocha` is available at https://gist.github.com/erikman/ecc095e3296f56fe37f3ecb712493cb7

please tell me if you need any other info, I'll see if I can do some debugging in the meantime :)

benrubson:
Does the target of the symlink exist ?
I think you may be facing #283, which has been corrected.
But has Mac OS does not support `llistxattr`, it should still experience the issue (see #305).

benrubson:
#342 should solve your issue 👍

benrubson:
Do not forget to run `make check` to be sure you pass all tests 👍

erikman:
Hi and thanks :)

No the symlink target does not exist (I'm copying a relative symlink).

#342 does not seem to solve the problem (same error), I've updated the logfile with encfs-benrubson-links.txt with the new log for same operation (`HEAD` at `ec9e94f0`).

`make check` fails on both branches (4/131). Log from `benrubson:links`:

```
Test project /Users/erik/projects/private/encfs/build
    Start 1: checkops
1/2 Test #1: checkops .........................   Passed    0.11 sec
    Start 2: scriptedtests
2/2 Test #2: scriptedtests ....................***Failed   18.68 sec
/Users/erik/projects/private/encfs/tests/normal.t.pl ... ok
xattr: No such file: /var/folders/q7/pd37snts1m156yr59y87w6fm0000gn/T//encfs-reverse-tests-je9_/decrypted/symlink

#   Failed test 'symlink to 'foo' extended attributes can be read (return code was 256)'
#   at /Users/erik/projects/private/encfs/tests/reverse.t.pl line 103.
#          got: '256'
#     expected: '0'
xattr: No such file: /var/folders/q7/pd37snts1m156yr59y87w6fm0000gn/T//encfs-reverse-tests-je9_/decrypted/symlink

[...]

PR #287, opened 2017-02-14
Resolve xattrs for symlinks (resolves #247, resolves #283)

dinoboy197:
Fix for #247 and #283.

PR #342, opened 2017-06-28
Make all xattr operations on link themselves

benrubson:
Hello,

This PR finishes works done in #287 making all xattr operations on links themselves instead of their targets.
It should also solve #341 using `XATTR_NOFOLLOW` Mac OS flag.

Thank you 👍 

Ben

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -93,20 +93,6 @@ check_cxx_source_compiles ("#include <sys/types.h>
   int main() { getxattr(0,0,0,0,0,0); return 1; }
   " XATTR_ADD_OPT)
 
-# Check if xattr function llistxattr exists.
-include (CheckCXXSourceCompiles)
-if (XATTR_ADD_OPT)
-check_cxx_source_compiles ("#include <sys/types.h>
-  #include <sys/xattr.h>
-  int main() { llistxattr(0,0,0,0); return 1; }
-  " XATTR_LLIST)
-else (XATTR_ADD_OPT)
-check_cxx_source_compiles ("#include <sys/types.h>
-  #include <sys/xattr.h>
-  int main() { llistxattr(0,0,0); return 1; }
-  " XATTR_LLIST)
-endif (XATTR_ADD_OPT)
-
 # Check if we have some standard functions.
 include (CheckFuncs)
 check_function_exists_glibc (lchmod HAVE_LCHMOD)
--- a/encfs/encfs.cpp
+++ b/encfs/encfs.cpp
@@ -660,7 +660,7 @@ int encfs_statfs(const char *path, struct statvfs *st) {
 #ifdef XATTR_ADD_OPT
 int _do_setxattr(EncFS_Context *, const string &cyName, const char *name,
                  const char *value, size_t size, uint32_t pos) {
-  int options = 0;
+  int options = XATTR_NOFOLLOW;
   return ::setxattr(cyName.c_str(), name, value, size, pos, options);
 }
 int encfs_setxattr(const char *path, const char *name, const char *value,
@@ -673,7 +673,7 @@ int encfs_setxattr(const char *path, const char *name, const char *value,
 #else
 int _do_setxattr(EncFS_Context *, const string &cyName, const char *name,
                  const char *value, size_t size, int flags) {
-  return ::setxattr(cyName.c_str(), name, value, size, flags);
+  return ::lsetxattr(cyName.c_str(), name, value, size, flags);
 }
 int encfs_setxattr(const char *path, const char *name, const char *value,
                    size_t size, int flags) {
@@ -686,7 +686,7 @@ int encfs_setxattr(const char *path, const char *name, const char *value,
 #ifdef XATTR_ADD_OPT
 int _do_getxattr(EncFS_Context *, const string &cyName, const char *name,
                  void *value, size_t size, uint32_t pos) {
-  int options = 0;
+  int options = XATTR_NOFOLLOW;
   return ::getxattr(cyName.c_str(), name, value, size, pos, options);
 }
 int encfs_getxattr(const char *path, const char *name, char *value, size_t size,
@@ -698,7 +698,7 @@ int encfs_getxattr(const char *path, const char *name, char *value, size_t size,
 #else
 int _do_getxattr(EncFS_Context *, const string &cyName, const char *name,
                  void *value, size_t size) {
-  return ::getxattr(cyName.c_str(), name, value, size);
+  return ::lgetxattr(cyName.c_str(), name, value, size);
 }
 int encfs_getxattr(const char *path, const char *name, char *value,
                    size_t size) {
@@ -710,12 +710,9 @@ int encfs_getxattr(const char *path, const char *name, char *value,
 
 int _do_listxattr(EncFS_Context *, const string &cyName, char *list,
                   size_t size) {
-#ifndef XATTR_LLIST
-#define llistxattr listxattr
-#endif
 #ifdef XATTR_ADD_OPT
-  int options = 0;
-  int res = ::llistxattr(cyName.c_str(), list, size, options);
+  int options = XATTR_NOFOLLOW;
+  int res = ::listxattr(cyName.c_str(), list, size, options);
 #else
   int res = ::llistxattr(cyName.c_str(), list, size);
 #endif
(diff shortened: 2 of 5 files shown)


rfjakob:
Hi Ben, thanks for the patch. It looks like Linux does not have XATTR_NOFOLLOW - is this handled in your code?

Also, I would really like to see a testcase in normal.t.pl added for this.

benrubson:
`XATTR_NOFOLLOW ` is only used if `XATTR_ADD_OPT` is defined, it will be added as the last option to xattr functions.
Linux does not define `XATTR_ADD_OPT`, so will not use `XATTR_NOFOLLOW` 👍

benrubson:
Regarding the test case, I think the one added in reverse.pl in #287 does the job ?

rfjakob:
Yes the test is a good improvement, but writing xattrs should also be
tested. If i saw it correctly the test only checks reading, right?

benrubson:
I think we're ready to merge @rfjakob 👍

benrubson:
@rfjakob let's commit this please ?
I'll then be able to re-arrange #338.
Thank you very much 👍

rfjakob:
Sorry, I will be focusing completely on #214 for now. This affects multiple users on multiple platforms. Debian 9 was just released that includes EncFS 1.9.1. We should give it all we got.

benrubson:
I totally agree, #214 must really be corrected.
If you plan to release 1.9.2, once #214 solved, I propose to come back to this (#342), as well as #295 which makes the code handling errors better.
This would make a nice 1.9.2 :)
Many thx @rfjakob 👍

mouse07410:
Could you tell me if there's any reason this PR hasn't been merged?

benrubson:
@mouse07410, @rfjakob who has the merge right is focusing on an even worst bug, #214.
Once solved he should review & merge this one :)

benrubson:
Perhaps you now have some time to see this tiny change @rfjakob ?
Many thx 👍

benrubson:
Shall we go with this one @rfjakob ?
Thank you very much for your support 👍

rfjakob:
Tests fail on my Fedora box:
```
$ g++ --version
g++ (GCC) 7.1.1 20170622 (Red Hat 7.1.1-3)

$ ./test.sh 
tests/normal.t.pl ... 
runTests: mode=standard
tests/normal.t.pl ... 4/122 attr_set: Operation not supported
Could not set "encfs" for /tmp/encfs-tests-iEas/crypt/data

#   Failed test 'extended attributes can be set (return code was 256)'
#   at tests/normal.t.pl line 324.
#          got: '256'
#     expected: '0'
attr_get: Operation not supported
Could not get "encfs" for /tmp/encfs-tests-iEas/crypt/data

#   Failed test 'extended attributes can be get (return code was 256)'
#   at tests/normal.t.pl line 326.
#          got: '256'
#     expected: '0'
attr_get: No data available
Could not get "encfs" for /tmp/encfs-tests-iEas/crypt/data-rel
tests/normal.t.pl ... 57/122 
runTests: mode=paranoia
tests/normal.t.pl ... 67/122 attr_set: Operation not supported
Could not set "encfs" for /tmp/encfs-tests-HoSG/crypt/data

#   Failed test 'extended attributes can be set (return code was 256)'
#   at tests/normal.t.pl line 324.
#          got: '256'
#     expected: '0'
attr_get: Operation not supported
Could not get "encfs" for /tmp/encfs-tests-HoSG/crypt/data

#   Failed test 'extended attributes can be get (return code was 256)'
#   at tests/normal.t.pl line 326.
#          got: '256'
#     expected: '0'
attr_get: No data available
Could not get "encfs" for /tmp/encfs-tests-HoSG/crypt/data-rel
tests/normal.t.pl ... 120/122 # Looks like you failed 4 tests of 122.
tests/normal.t.pl ... Dubious, test returned 4 (wstat 1024, 0x400)
Failed 4/122 subtests 
	(less 2 skipped subtests: 116 okay)
tests/reverse.t.pl .. ok     

Test Summary Report
-------------------
tests/normal.t.pl (Wstat: 1024 Tests: 122 Failed: 4)
  Failed tests:  18-19, 81-82
  Non-zero exit status: 4
Files=2, Tests=147, 23 wallclock secs ( 0.04 usr  0.01 sys + 20.10 cusr  0.39 csys = 20.54 CPU)
Result: FAIL
Failed 1/2 test programs. 4/147 subtests failed.
```

benrubson:
Does your `/tmp` have `user_xattr` ?

rfjakob:
Fedora mounts a tmpfs to /tmp, and it looks like tmpfs does not support user_xattr:
```
0 jakob@brikett:/tmp$ attr -s user.abc -V def foo
attr_set: Operation not supported
Could not set "user.abc" for foo
```
This means xattr testing must move to /var/tmp

benrubson:
Do you think we should modify this in both pl test scripts ?
`my $tempDir = $ENV{'TMPDIR'} || "/tmp";`
to
`my $tempDir = "/var/tmp";`
?

rfjakob:
I have limited it do Linux like this: https://github.com/vgough/encfs/pull/356/commits/9a169bfb78f4249120ea971aaff5e5a8307b5697#diff-6b2afe0bcb135d3fd0af455eb74bdd65R15

Looks better now, however, I get some `Could not get ...` messages:
```
$ ./test.sh 
tests/normal.t.pl ... 
runTests: mode=standard
tests/normal.t.pl ... 1/122 attr_get: No data available
Could not get "encfs" for /var/tmp/encfs-tests-25n0/crypt/data-rel
tests/normal.t.pl ... 55/122 
runTests: mode=paranoia
tests/normal.t.pl ... 69/122 attr_get: No data available
Could not get "encfs" for /var/tmp/encfs-tests-tvlS/crypt/data-rel
tests/normal.t.pl ... ok       
tests/reverse.t.pl .. ok     
All tests successful.
Files=2, Tests=147, 22 wallclock secs ( 0.05 usr  0.00 sys + 20.18 cusr  0.39 csys = 20.62 CPU)
Result: PASS
```

rfjakob:
Yes, these also appear in Travis: https://travis-ci.org/vgough/encfs/jobs/258947994#L852

Why doesn't "make check" report this?

benrubson:
The `Could not get ...` messages are expected :
```
$return_code = ($have_xattr) ? system(@getattr, "$crypt/data-rel") : 1;
isnt($return_code, 0, "extended attributes operations do not follow symlinks (return code was $return_code)");
```
On Linux, we can't set extended attributes on symlinks (FreeBSD allows this).
So to check that symlinks are not followed when attributes are set/get, this test checks that it does not get the attribute of the link's target :)

`make check` only outputs the failed tests (proper / easier to read output).

benrubson:
> `make check` only outputs the failed tests (proper / easier to read output).

Thanks to `--output-on-failure` in `CMakeLists.txt`.

benrubson:
I think the following would then need to be updated due to the `/var/tmp` tweak :
https://github.com/vgough/encfs/blob/master/test.sh#L7
`grep -E "(/var)?/tmp/encfs-reverse-tests-|(/var)?/tmp/encfs-tests-"`

Regarding `Could not get ...` messages, they are "expected" :
```
$return_code = ($have_xattr) ? system(@getattr, "$crypt/data-rel") : 1;
isnt($return_code, 0, "extended attributes operations do not follow symlinks (return code was $return_code)");
```
On Linux, we can't set extended attributes on symlinks (FreeBSD allows this).
So to check that symlinks are not followed when attributes are set/get, this test checks that it does not get the attribute of the link's target :)

**I then think we should add `2>/dev/null` to this specific test.**

Regarding `make check`, it only outputs the failed tests (proper / easier to read output, especially when they are pasted into bug reports).
So this thanks to `--output-on-failure` in `CMakeLists.txt`.

**I then think `./test.sh` here is redundant :**
https://github.com/vgough/encfs/blob/master/.travis.yml#L28

Last thing I noted (not really related to this PR) :
Here we have verbose :
https://github.com/vgough/encfs/blob/master/CMakeLists.txt#L272
Here we have debug :
https://github.com/vgough/encfs/blob/master/test.sh#L16
Not sure which one we should adopt, but we could standardize :)

Thank U Jakob 👍

rfjakob:
@benrubson This --output-on-failure` thing does not seem to work properly. Check out https://github.com/vgough/encfs/issues/358 , there is zero useful info in the output.

benrubson:
The command to use to have the failed output is `make check` as now described in `INSTALL.md` file :)
Not `make test` as @svetlemodry used in #358 :)

rfjakob:
Ok. Why do we have both?

benrubson:
Unfortunately you can't modify CMake test target to output errors.
This is why we have our own check target.
I added it when I added the Perl tests to the CMake file (#309), to be able to run everything with the make command :)

rfjakob:
Hmm that's too bad and pretty confusing for users. Maybe we should drop both `make test` and `make check` and have just `test.sh`. People will keep running `make test` all the time!

rfjakob:
Merged as https://github.com/vgough/encfs/pull/356

PR #357, opened 2017-07-29
travis ci: enable gcc, call build.sh && test.sh

rfjakob:
gcc is default compiler on most distributions, we should
also test with it.

build.sh exists and should be tested as well, so why not use
it in Travis.
test.sh outputs much more details than "make check", so
run it as well.

Additionally, make build.sh and test.sh work when called from
other directories.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -1,15 +1,15 @@
-#!/bin/bash
+#!/bin/bash -eu
 
-set -eu
+# Make sure we are in the directory this script is in.
+cd "$(dirname "$0")"
 
-if [ ! -d build ]
+if [[ ! -d build ]]
 then
 	mkdir build
 	cd build
 	cmake ..
-else
-	cd build
+	cd ..
 fi
 
-make
+make -j2 -C build
 
--- a/test.sh
+++ b/test.sh
@@ -1,10 +1,16 @@
 #!/bin/bash -eu
 
-./build/checkops &> /dev/null
+# Make sure we are in the directory this script is in.
+cd "$(dirname "$0")"
 
+# Failed tests can leave dangling mounts behind.
 for i in $(mount | grep -e "/tmp/encfs-reverse-tests-\|/tmp/encfs-tests-" | cut -f3 -d" "); do
 	echo "Warning: unmounting leftover filesystem: $i"
-	fusermount -u $i
+	fusermount -u $i || true
 done
+
+# This is very noisy so run it silently at first. Run it again with
+# output if the first run fails.
+./build/checkops &> /dev/null || ./build/checkops
 
 perl -MTest::Harness -e '$$Test::Harness::debug=1; runtests @ARGV;' tests/*.t.pl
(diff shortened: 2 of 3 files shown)


build.sh
```diff
@@ -1,15 +1,15 @@
...
 	mkdir build
 	cd build
 	cmake ..
-else
-	cd build
+	cd ..
 fi
 
-make
+make -j2 -C build
```
samrocketman: Perhaps it's a good idea to set `-j` to `nproc`.

test.sh
```diff
@@ -1,10 +1,16 @@
 #!/bin/bash -eu
 
-./build/checkops &> /dev/null
+# Make sure we are in the directory this script is in.
+cd "$(dirname "$0")"
```
samrocketman: See former `dirname` in a subshell comment.

test.sh
```diff
@@ -1,10 +1,16 @@
...
 
+# Failed tests can leave dangling mounts behind.
 for i in $(mount | grep -e "/tmp/encfs-reverse-tests-\|/tmp/encfs-tests-" | cut -f3 -d" "); do
 	echo "Warning: unmounting leftover filesystem: $i"
-	fusermount -u $i
+	fusermount -u $i || true
```
samrocketman: Why swallow the exit code?  What if there's a legitimate error?

build.sh
```diff
@@ -1,15 +1,15 @@
-#!/bin/bash
+#!/bin/bash -eu
 
-set -eu
+# Make sure we are in the directory this script is in.
+cd "$(dirname "$0")"
```
samrocketman: Rather than `dirname` in a subshell, I prefer bash parameter expansion: `${0%/*}`.

build.sh
```diff
@@ -1,15 +1,15 @@
...
 
-if [ ! -d build ]
+if [[ ! -d build ]]
```
samrocketman: `[` test command did the same thing.  Not really against this; just pointing it out.

build.sh
rfjakob: I think in this case, clarity is more important than brevity or speed. I mean, look at it, `${0%/*}`? It's horrible!

build.sh
rfjakob: The bash variant `[[` is usually safer than the sh variant `[`. It does not matter here, but I try to consistently use `[[`.

build.sh
rfjakob: Thought about it, but `nproc` is not available on MacOS!

test.sh
rfjakob: An error here should not keep the test suite from running. We may get `device or resource busy` or even `command not found` (on MacOS). Tests should still be run.

build.sh
samrocketman: I can see your argument.  I tend to pair it with a comment describing it is bash Parameter Expansion from the man page.

PR #360, opened 2017-07-31
Correct build env

benrubson:
Hi,

This PR :
- adds a comment to `FindIntl.cmake` file to say why it is there ;
- removes temporary directories left by dangling mounts ;
- makes `test.sh` FreeBSD (and others) compatible (`fusermount` may not exist).

Thank you 👍 

Ben

### DIFF
--- a/cmake/FindIntl.cmake
+++ b/cmake/FindIntl.cmake
@@ -1,3 +1,8 @@
+# This file comes from https://github.com/Kitware/CMake/blob/master/Modules/FindIntl.cmake
+# It is needed for cmake < 3.2.0 which does not have it.
+# When cmake 3.0.2 (Debian jessie version) will really be outdated, we will be able to
+# delete this file and bump cmake version in CMakeLists.txt and INSTALL.md.
+
 # Distributed under the OSI-approved BSD 3-Clause License.  See accompanying
 # file Copyright.txt or https://cmake.org/licensing for details.
 
--- a/test.sh
+++ b/test.sh
@@ -4,15 +4,24 @@
 cd "$(dirname "$0")"
 
 # Failed tests can leave dangling mounts behind.
-for i in $(mount | grep -e "/tmp/encfs-reverse-tests-\|/tmp/encfs-tests-" | cut -f3 -d" "); do
+todel+=("")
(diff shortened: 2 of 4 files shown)


rfjakob:
I disagree here. Dangling mounts should not be purged *after* the tests because that means you can no longer investigate what went wrong.

The freebsd unmount thing adds a horrible amount of complexity (that sed line!) that i don't want to have in this simple script. Let's just wrap it in an "$OSTYPE is linux" check so bsd users dont get error messages.

benrubson:
OK, I then just reverted order back to origin 👍 

Not sure Linux is the only one to have `fusermount`, this is why I used `which` to check if it is there or not. Perhaps we should simply use `umount` for all OS ?

The `sed` line is for another purpose, it extracts the working directory from the mount point so that it can then be deleted later (without this dangling working directories are left / are not deleted).

benrubson:
I modified to use `umount` for all OS which tends to make the script simpler, I also think it's better.

benrubson:
Does it sounds good to you like this Jakob ?
Thank you 👍

rfjakob:
Uhmm, have you tested this on Linux?
```
$ umount b
umount: b: Permission denied
$ fusermount -u b
(ok)
```

benrubson:
Not with a non-root user :)
OK, we then need to use `fusermount` when available.

benrubson:
OK, `fusermount` back in the race :)
I know you're not so found of adding so many lines, but the `which` solution seems to be the cleaner one.
We could also do `fusermount -u $i || umount -f $i || true`, but it would produce error messages on machines without `fusermount`.

rfjakob:
I think you are right that it's no good to have functionality in test.sh that does not work on FreeBSD and MacOS, but at this point I think it's better to just drop the function completely than to add more complexity to `test.sh`

rfjakob:
Obsoleted by https://github.com/vgough/encfs/commit/22a88c0e489b2b4abc6480d5690196fa5c217948

benrubson:
Thank you for your commit Jakob.

benrubson:
We could create a `test_force_unmount.sh` with the neede piece of code.

rfjakob:
I don't think we really need that functionality, let's not add unneccessary code

PR #362, opened 2017-08-01
Say why FindIntl.cmake is there

benrubson:
Hi,
This PR simply says why `cmake/FindIntl.cmake` is there.
Thx !


rfjakob:
I'd rather keep this file in sync with upstream

benrubson:
OK :)
Note that it is slightly modified, the following line had to be commented :
`include(${CMAKE_CURRENT_LIST_DIR}/FindPackageHandleStandardArgs.cmake)`

PR #363, opened 2017-08-04
add clang-tidy build support

vgough:
Requires CMAKE 3.6, as that's when CLANG_TIDY macros appear to have been added: https://cmake.org/cmake/help/v3.6/release/3.6.html#properties


CMakeLists.txt
```diff
@@ -233,5 +249,5 @@ if (POD2MAN)
     DESTINATION share/man/man1)
 endif (POD2MAN)
 
-add_custom_target(test COMMAND ${CMAKE_CURRENT_LIST_DIR}/test.sh)
+add_custom_target(tests COMMAND ${CMAKE_CURRENT_LIST_DIR}/test.sh)
```
rfjakob: Is this intentional?

vgough:
Updated INSTALL.md, thanks Ben.

PR #367, opened 2017-08-05
Use word tests in plural everywhere

benrubson:
Hi,

This PR follows commit https://github.com/vgough/encfs/commit/fea77bad4e3366f0e5b2317f26c7405a896a4480 where `test` target has been renamed to `tests`.
Let's standardize the word _tests_ :)

Thx 👍 

Ben


benrubson:
Or we could also revert `tests` target to `test` :)
Up to U !

vgough:
Yeah, the names are confusing.  I'm adding unit tests, which cmake will run with "make test".  So the integration tests need a better name.  Is there a standard name for an integration test target?

benrubson:
What about the following ?

`make test` and `make tests` would simply display the following :
- _use `make unitests` to run unity tests (developers only ?) ;_
- _use `make intests` to run integration tests (users)._

Not sure about the other word we could use.
There is `check`,  but as @rfjakob mentioned earlier, users, are used to run `make test`, so I think we should somehow make them run the correct tests from this command :)

vgough:
Naively, I'd like for 'make test' to run unit tests, since it seems like that's a common expectation when a cmake project has tests enabled.  I know that my IDE will show and parse test results from the test target.  I don't use cmake day to day, so I'm willing to learn how to do it correctly if I'm not.

benrubson:
I don't either use cmake on a day to day basis, so not sure what is the best practice here :|

> I know that my IDE will show and parse test results from the test target.

Perhaps then we could simply keep the `test` target as you began to use it, and simply add `/test.sh` to it as a last test ?
Everything would then be run by every user compiling this project (and using `make test`).

This would of course however increase tests running time in your IDE...

vgough:
I renamed targets, leaving "test" pointing to unit tests for development, and "integration" for running just the live filesystem integration tests.  Thanks for the feedback.

PR #368, opened 2017-08-05
Syslog options

benrubson:
Hi,

This PR adds the PID of the process to syslog logging.
As well as explicity choosing user facility, which is the default.

Thank you 👍 

Ben


vgough:
lgtm, thanks.

PR #369, opened 2017-08-06
use git-vendor to track bundled dependencies

vgough:
Updates tinyxml2 and easyloggingc++, letting git-vendor track versions and sources.  See README.md in vendor subdirectory.  Internal subdirectory is removed.


vgough:
Sure, I'll pin to tags v9.95.0 and 5.0.1 respectively.

PR #373, opened 2017-08-07
add unit test and benchmark targets

vgough:
Both targets behind CMake flag, so they can be disabled.

Vendors google test library, start converting checks into unit tests which will be run by CMake's "test" target.

Vendors google benchmarking library, adds a simple first benchmark of memory pool allocation.  Built as "benchmarks" executable.


benrubson:
What about a `unitests` subfolder for all test files ?
(and then renaming `/test.sh` and `/tests` to `/intests.sh` and `/intests` ?)
(also see #367 where we are thinking about correct names to choose)

So if I understand correctly, `/encfs/test.cpp` (`checkops `) will disappear ?

Last question, why choosing circleci over travis for running your new tests ?
(still not sure why we keep circleci, as it is much less configured than travis and does not even run `/test.sh`)

Thx 👍

vgough:
Thanks for feedback. 

I've moved the tests to a "test" directory, and renamed the former test to 'integration' since it's doing an integration test with actual filesystem mounts.

So 'make test' will now run the unit tests.  And yes, checkops should eventually go away in favor of unit test coverage.

benrubson:
Perfect, many thx 👍 
I just noticed Travis does not run `make integration`.
Perhaps adding it to `ci/check.sh` ?

What about circleci ? It does not even run anymore when we push / update a PR.

Thx again 👍

PR #380, opened 2017-08-09
Code sanity check

benrubson:
Hello,

This PR follows #295 which has been lost.
I recovered all work in this new PR (thanks to github API).

Let's continue here then (sorry for this).

Many thanks 👍


encfs/RawFileIO.cpp
```diff
@@ -159,7 +158,7 @@ int RawFileIO::open(int flags) {
       result = fd = newFd;
     } else {
       result = -errno;
```
vgough: eno here?

encfs/RawFileIO.cpp
```diff
@@ -220,14 +221,17 @@ bool RawFileIO::write(const IORequest &req) {
   ssize_t bytes = req.dataLen;
   off_t offset = req.offset;
 
+  int eno = 0;
   while ((bytes != 0) && retrys > 0) {
+    errno = 0;
```
vgough: think my earlier comments were lost.
I'd rather not mutate errno if not necessary, and I don't think you need to do that here.

ie:
```
while ((bytes != 0 ...) {
  auto writeSize = ::pwrite(...);
  if (writeSize < 0) {
     auto eno = errno;
     ...
     return -eno;
  }
  ...

  knownSize = false;
  return -EIO;
}
```

encfs/RawFileIO.cpp
```diff
@@ -240,13 +244,16 @@ bool RawFileIO::write(const IORequest &req) {
...
                 << req.dataLen << ", max retries reached";
     knownSize = false;
-    return false;
-  }
-  if (knownSize) {
-    off_t last = req.offset + req.dataLen;
-    if (last > fileSize) {
-      fileSize = last;
+    return (eno) ? -eno : -EIO;
+  } else {
```
vgough: clang-tidy will complain about this.  No need to have an `else` clause after a block that returns.

Please remove the else bracket,
 ie `} else { if (knownSize) ...` -> `} \n if (knownSize) ...`

encfs/RawFileIO.cpp
benrubson: I just wondered why `write` would not have succeeded after having written something 10 times...
So I thought, as it did not manage to write what we want, perhaps it wrote only a small piece of our data and gently set `errno` to explain why.
Of course, not sure however it can happen...

Do you have a situation in mind where `write` could succeed 10 times without writing all the data ?

encfs/RawFileIO.cpp
benrubson: Corrected !

encfs/RawFileIO.cpp
benrubson: Yes will look better, corrected !

encfs/RawFileIO.cpp
```diff
@@ -219,37 +219,44 @@ ssize_t RawFileIO::write(const IORequest &req) {
...
-  while ((bytes != 0) && retrys > 0) {
-    errno = 0;
+  /*
+   * Let's write while pwrite() writes, to avoid writing only a part of the request,
+   * whereas it could have been fully written. This to avoid inconsistencies / corruption.
+   */
+  // while ((bytes != 0) && retrys > 0) {
+  while (bytes != 0) {
     ssize_t writeSize = ::pwrite(fd, buf, bytes, offset);
     eno = errno;
```
vgough: The changes here did not fix errno handling.  This change is making matters worse IMO.  pwrite() == 0 is not a problem.  Errno is still being mutated and handled improperly.

benrubson: I just pushed new changes @vgough, should be OK now.
Regarding the `errno = 0`, I explain it in a comment :
_just to be sure error seen in integration tests really comes from the function rc._

benrubson:
I've finished with this PR, so if it's OK for you @vgough, we can merge 👍 
Thank you !

vgough:
I've made a few changes, mostly getting rid of errno mutation and fix clang-tidy warnings, then running clang-format.  Closing this, as all your commits will appear in #391.  

Thanks!

PR #385, opened 2017-08-15
Make Cirle run all tests

benrubson:
Hi,

This PR makes Circle run all tests, as Travis does.
It also makes Travis run clang-tidy.
Travis will use clang, Circle gcc, so that both compilers are tested.

Thx 👍 

Ben

### DIFF
--- a/ci/setup.sh
+++ b/ci/setup.sh
@@ -1,10 +1,18 @@
 #!/bin/bash -eu
 
-if [ "$TRAVIS_OS_NAME" == "linux" ]; then
+if uname -s | grep -q Linux; then
   sudo modprobe fuse
+  # Download cmake > 3.5 so that we can run clang-tidy
+  # Circle will use an older one (2.8 in Trusty)
+  if [ "$TRAVIS" == "true" ]; then
+    wget https://cmake.org/files/v3.9/cmake-3.9.1-Linux-x86_64.tar.gz -O /tmp/cmake.tar.gz
+    tar -C /tmp/ -xf /tmp/cmake.tar.gz
+    sudo rm -f $(which cmake)
+    sudo ln -s $(ls -1 /tmp/cmake*/bin/cmake) /bin/
+  fi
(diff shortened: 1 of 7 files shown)


benrubson:
To speed-up builds, as Travis also already has osx to build ?
I think this is why @vgough made Travis test only one compiler a few days ago.

As there is also CircleCI configured, I thought it could be good to make it test another compiler, so one compiler per CI.
But of course we can change this behaviour :)

benrubson:
We could also make Travis run both clang and gcc, and remove Circle.
It would make things easier.
The issue with Circle is that it does not run when we submit a PR, it only runs where PR are merged, so we don't have its result as soon as we submit the PR.

ci/setup.sh
```diff
@@ -1,10 +1,18 @@
 #!/bin/bash -eu
 
-if [ "$TRAVIS_OS_NAME" == "linux" ]; then
+if uname -s | grep -q Linux; then
   sudo modprobe fuse
```
vgough: This doesn't work in circle, as it can't load fuse in the container.

vgough:
I don't think that circle can run the integration tests - at least not the configuration we have today which uses a container AFAIK.  I've imported these changes into a branch, but have to make some changes to work on circle.

benrubson:
Why do we keep Circle then ?
Let's give Travis all the jobs ?
It will also be easier to maintain, don't you think so ?

ci/setup.sh
benrubson: Mmmh yes according to the Circle build log, some kernel files are missing.
Re-installating (`apt-get`) the kernel or reconfiguring it (`dpkg-reconfigure`) could help, but these are quite uncommon build steps :)

vgough:
I've gone ahead and removed the circle config in my branch.

vgough:
@benrubson : I've used your work here to make #389.  Please take a look.

PR #389, opened 2017-08-20
drop circle, add clang-tidy to travis build

vgough:
The build.sh script is now used by CI.  It defaults to enabling integration tests, but not clang-tidy.  The Travis configuration enables the following:

Linux: clang + clang-tidy + integration tests
Linux: gcc + unit tests
OSX: clang + integration tests

Based on Ben's work in #385


ci/setup.sh
```diff
@@ -1,10 +1,23 @@
...
+  elif uname -s | grep -q Darwin; then
+    brew cask install osxfuse
+  fi
 fi
 
-if [ "$TRAVIS_OS_NAME" == "osx" ]; then
-  brew cask install osxfuse
+if [[ "$CHECK" == "true" ]]; then
+  if uname -s | grep -q Linux; then
+    wget https://cmake.org/files/v3.9/cmake-3.9.1-Linux-x86_64.tar.gz -O /tmp/cmake.tar.gz
```
benrubson: Perhaps we could add a condition here, so that when we will have cmake >3.6 built-in, we will not more download it :
`if ! which cmake >/dev/null 2>&1 || [[ $(cmake --version | awk '{print $3}') < 3.6 ]]`
Up to you 👍

benrubson:
Really nice build scripts 👍 
LGTM 👍 
I added one optional comment :)
Thx !

ci/setup.sh
benrubson: Even without the first part it works :
`if [[ $(cmake --version 2>/dev/null | awk '{print $3}') < 3.6 ]]`

benrubson: _just updated the condition above_

vgough: hmm, since this script is only used w/ CI, I think we should just remove this when/if the travis image we use comes with a later cmake.

benrubson: Perfect 👍

benrubson:
Strangely enough @vgough, when PR are merged, sounds like Circle is still running :

screen shot 2017-08-27 at 21 00 27

vgough:
Thanks, there was still a webhook configured.  I've removed it.

PR #391, opened 2017-08-26
Return code rework

vgough:
This takes #380, eliminates mutation of errno, fixes formatting and fixes clang-tidy warnings.


encfs/main.cpp
```diff
@@ -532,10 +532,9 @@ void *encfs_init(fuse_conn_info *conn) {
     int res =
         pthread_create(&ctx->monitorThread, nullptr, idleMonitor, (void *)ctx);
     if (res != 0) {
-      int eno = errno;
       RLOG(ERROR) << "error starting idle monitor thread, "
                      "res = "
-                  << res << ", " << strerror(eno);
+                  << res << ", " << strerror(res);
     }
```
benrubson: Good catch, `pthread_create` does not set `errno` 👍

encfs/RawFileIO.cpp
```diff
@@ -225,23 +226,25 @@ ssize_t RawFileIO::write(const IORequest &req) {
...
-      if (eno) {
-      	return -eno;
-      }
+      // pwrite is not expected to return 0, so eno should always be set, but we
+      // never know...
+      return -eno;
+    }
+    if (writeSize == 0) {
       return -EIO;
     }
```
benrubson: Here we return `EIO` we loose a chance to return correct `errno` : depending on implementations `pwrite` could set `errno` when it returns 0.
This is why I tested `if (writeSize <= 0)` [in my initial commit](https://github.com/vgough/encfs/blob/acbeb647658a334ff00b2f2b1dac2addb2082bad/encfs/RawFileIO.cpp#L235).
Could we then get back to previous version here ?
Many thanks 👍

encfs/RawFileIO.cpp
```diff
@@ -206,8 +209,6 @@ ssize_t RawFileIO::read(const IORequest &req) const {
 
   if (readSize < 0) {
     int eno = errno;
-    errno = 0;  // just to be sure error seen in integration tests really comes
-                // from the function rc.
```
benrubson: I would have been pleased to keep this, as explained in the comment just to be sure...
Thank you 👍

PR #393, opened 2017-08-26
Run "sudo" tests also if user is root

benrubson:
Hello,

This PR enables "sudo" tests if integration tests are run using root user,
which can be convenient to be sure to run all tests.

Thank you 👍 

Ben

### DIFF
--- a/integration/normal.t.pl
+++ b/integration/normal.t.pl
@@ -34,11 +34,23 @@ if(system("which lsextattr > /dev/null 2>&1") == 0)
     $setattr = "setextattr -h user encfs hello";
     $getattr = "getextattr -h user encfs";
 }
+
 # Do we support xattr ?
 my $have_xattr = 1;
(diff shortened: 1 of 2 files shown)


integration/normal.t.pl
```diff
@@ -34,13 +34,25 @@
...
 # Do we support xattr ?
 my $have_xattr = 1;
 if(system("./build/encfs --verbose --version 2>&1 | grep -q HAVE_XATTR") != 0)
 {
     $have_xattr = 0;
 }
 
+# Did we ask, or are we simply able to run "sudo" tests ?
+my $sudo_cmd;
+if (! $<)
```
vgough: should this be `$>` to check effective user id rather than real user id?

Please make the test condition explicit, eg `if ( $> != 0 ) {`

benrubson: Updated 👍

PR #395, opened 2017-08-28
Print whether or not we are running sudo tests

benrubson:
Hi,

This change only prints a flag to know whether or not we are running sudo tests.

Thx 👍


benrubson:
_I merged it myself as it's quite tiny. Thx !_

PR #396, opened 2017-08-31
Document cloud usage in man page

### LINKED ISSUE #141: Add cloud-safe configuration option
We have an encrypted folder on an OwnCloud server which is shared by 2 users. If user1 moves a file from a subfolder to another, user2 gets an I/O error when accessing this file. User1 has no problem with this file. If user1 copies this file, there is no error.Without encfs there was no problem when moving files.

Regards, Jan

benrubson:
Hi,

Here are some comments in man page to discourage using external IV chaining with cloud providers.
We should then be able to somehow close #141.

Thx 👍 

Ben

### DIFF
--- a/README.md
+++ b/README.md
@@ -82,7 +82,8 @@ How about a nice email instead?
 
 ### What settings should I use for Dropbox?
 
-Use **standard mode**. There have been [reports](https://github.com/vgough/encfs/issues/388)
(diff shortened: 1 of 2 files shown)


benrubson:
_I merged it myself as it is just related to documentation. Thx !_

PR #401, opened 2017-09-30
Ensure compatibility with LibreSSL

vaygr:
This fixes compilation with [LibreSSL](https://www.libressl.org/). Tested with **2.2.9+**.


benrubson:
Many thanks for the patch 👍 
Are these `define` also needed for last versions of LibreSSL (2.5.5 / 2.6.2) ?

benrubson:
After some researches, sounds like yes.

Issue #328, opened 2017-06-03
Increase FUSE_USE_VERSION ?


benrubson:
Hi,

Is there any reason to keep `FUSE_USE_VERSION` to `26` ?
https://github.com/vgough/encfs/blob/v1.9.1/CMakeLists.txt#L61

What about increasing it to last 2.x series, `29` ?
Any benefits ? Any drawbacks ?

(_Debian Jessie has libfuse 2.9_)

Many thx 👍 

Ben

l29ah:
Drawback: the users of the old libfuse won't be able to use it.

benrubson:
Thnak you for your feedback Sergey 👍 
Even Debian Wheezy (oldoldstable, may 2013) has libfuse 2.9.
But perhaps you personally use an older version or think about systems having it ?
Thank you !

benrubson:
@vgough, any thoughts ?
Thx 👍

vgough:
Linux isn't a problem, we could probably bump significantly.  I'm more worried about non-Linux FUSE implementations.  Digging into OSXFuse, it looks like it uses 26 API: https://github.com/osxfuse/fuse/blob/master/include/fuse.h#L22

If the benefits are big enough, we could probably make encfs work with either 26 or later APIs.  I don't really know what the benefits are.

benrubson:
I'm not sure either what are the benefits, perhaps we should ask libfuse guys.
I did not notice OSXFuse was stuck to 26 API.
Of course Linux has 2.9 since a long time, FreeBSD too.

benrubson:
Here's at least the changelog :

```
What is new in 2.9

 - Add "zero copy" support for kernel 2.6.35 or newer

 - Make maximum background requests tunable on kernel 2.6.32 or newer

 - Require --no-canonicalize in (u)mount (util-linux version 2.18 or
   newer) to fix security problems with fusermount

 - Use dynamically sized hash tables in high level library

 - Memory use of filesystem daemon can shrink more easily

 - Add "auto_unmount" option

 - Add "remember" option

 - Add man pages for fusermount, mount.fuse and ulockmgr_server

 - API changes:

    o Introduce "store" and "retrieve" for accessing kernel buffers on
      kernel 2.6.36 or newer

    o Introduce abstract buffer for zero copy operations

    o Allow path calculation to be omitted on certain operations

    o Allow batching forget requests

    o Add "flock" method

    o Add support for ioctl on directories

    o Add delete notification

[...]

PR #404, opened 2017-10-01
Bump FUSE_USE_VERSION to 29

benrubson:
Hi,

Related to #328, let's test bumping FUSE_USE_VERSION to 29 to see if all our CI tests run correctly (especially the Mac OS one).

Ben


benrubson:
OK so tests pass correctly, we could then perhaps think about merging this in order to use the last 2.x Fuse lib series.

benrubson:
@rfjakob any thoughts on this ?
Sounds quite reasonable to me (as mentioned in #328, even Debian Wheezy (oldoldstable, may 2013) has libfuse 2.9).
vgough seems to be OK according to his comment there.
Thx 👍

rfjakob:
I'm a bit worried about osxfuse. Their implementation has always been way more brittle, and maybe there is a reason they kept their default at version 26?

benrubson:
I've asked @bfleischer, [osxfuse](https://github.com/osxfuse/osxfuse) maintainer, to give us his feedback 👍

bfleischer:
FUSE for macOS 3.x has been shipping with lib(osx)fuse 2.9.x since May 2015. I'm not aware of compatibility issues.

Defining `FUSE_USE_VERSION` to 29 instead of 26 should make no difference at all. If I have not overlooked something there is not even a version check in libfuse that differentiates between version 26 and 29.

> I'm a bit worried about osxfuse. Their implementation has always been way more brittle, and maybe there is a reason they kept their default at version 26?

What do you mean by "brittle"? FUSE for macOS ships with two versions of the FUSE library:

* libosxfuse 2.9.7: This is the default FUSE library on macOS. I know the renaming is not ideal, but it was necessary to keep binary compatibility with old MacFUSE file systems.
* libfuse 2.7.4: This is the MacFUSE compatibility layer needed to support legacy MacFUSE file systems. It is only available if the user selected this option during the install. No new file systems should be linked against this library.

benrubson:
> FUSE for macOS 3.x has been shipping with lib(osx)fuse 2.9.x since May 2015.

This is the one we use, at least in our CI environment :
```
brew cask install osxfuse
(...)
Downloading https://github.com/osxfuse/osxfuse/releases/download/osxfuse-3.5.8/osxfuse-3.5.8.dmg
```

Perfect, thank you very much for your comment @bfleischer, really appreciated 👍

samrocketman:
I realize I'm a bit late to this discussion and I do not recommend chsnging history.  However, going forward I would prefer changes like this to be rebased.  6 commits to change one line is a bit much.  Rebasing would have discarded all of the merge commits.  Just some food for thought on my workflow preference.

benrubson:
"Squash & merge" has been used to make master history clean, keeping only one global entry 👍

PR #416, opened 2017-10-09
Correctly sort headers

benrubson:
to get rid of llvm-include-order warnings.


benrubson:
Will need to squash & merge to make master log clean :)

PR #417, opened 2017-10-09
Correct  an undefined pointer dereference

benrubson:
Hi,

This PR corrects an undefined pointer dereference detected by clang.

Ben

### DIFF
--- a/encfs/readpassphrase.cpp
+++ b/encfs/readpassphrase.cpp
@@ -143,8 +143,8 @@ restart:
         *p++ = ch;
       }
     }
+    *p = '\0';
   }
-  *p = '\0';
   save_errno = errno;
   if ((term.c_lflag & ECHO) == 0u) {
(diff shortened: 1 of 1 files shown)


vgough:
Wonder if we shouldn't update this file entirely.  This version is from 2001, so maybe we should pull in something that has been developed further, eg https://github.com/openssh/openssh-portable/blob/master/openbsd-compat/readpassphrase.c

benrubson:
Sure, if you think it's worth it, then feel free to close this PR 👍

benrubson:
Not sure it's really worth it, files are almost the same.
Undefined pointer dereference is due to an older PR which checks the write return code a few lines above, and the faulty line was then wrongly non-enclosed.
This may not arrive anymore when we will have corrected all compilation warnings & clang-tidy messages, we will be able to easily check that PR do not generate new messages before merging 👍

PR #418, opened 2017-10-09
Correct an uninitialized value

benrubson:
Hi,

Trying to correct this from clang :
```/home/travis/build/vgough/encfs/encfs/ConfigVar.cpp:197:5: warning: Function call argument is a pointer to uninitialized value```

Ben


benrubson:
Ok it solves the issue, I'll however wait for your approval @vgough.

Issue #420, opened 2017-10-10
Configuration fails if CXXFLAGS set


mouse07410:
MacOS Siera 10.12.6, Xcode-9.0, ton of packages installed via Macports into `/opt/local` - including `OpenSSL-1.0.2l` and `gnuregex-0.12.2`.

Problems:

1. Build process does not figure to look for installed packages in `/opt/local`.
2. Build process fails to find `gnuregex` installed by Macports in `/opt/local`.
3. Build process stitches compiler flags from CXXFLAGS and whatever CMake cares to add using ";", ending up with:

```
. . . . .
Building CXX object CMakeFiles/cmTC_f7efa.dir/src.cxx.o
/opt/local/bin/clang++    -maes -mpclmul -mrdrnd -msse2 -mssse3 -msse4.2 -mtune=native -Os -Ofast -std=c++11;-Wall -Wextra -DHAVE_CXX_FLAG_COVERAGE  --coverage   --coverage -o CMakeFiles/cmTC_f7efa.dir/src.cxx.o -c /Users/uri/src/encfs/build/CMakeFiles/CMakeTmp/src.cxx
clang: error: no input files
/bin/sh: -Wall: command not found
```

Would be great to have this resolved.

**Update**
This is the ugly hack I had to apply to make it working (but once applied - all the tests passed ;).
```
diff --git a/CMakeLists.txt b/CMakeLists.txt
index 8ae405e4..a23caeec 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -37,17 +37,18 @@ if (${CMAKE_MAJOR_VERSION}.${CMAKE_MINOR_VERSION} GREATER 3.0)
 else ()
   if (CMAKE_COMPILER_IS_GNUCXX)
     message ("** Assuming that GNU CXX uses -std=c++11 flag for C++11 compatibility.")
-    list(APPEND CMAKE_CXX_FLAGS "-std=c++11")
+    list(APPEND CMAKE_CXX_FLAGS " -std=c++11 ")
   elseif (CMAKE_CXX_COMPILER_ID MATCHES "Clang")
     message ("** Assuming that Clang uses -std=c++11 flag for C++11 compatibility.")
-    list(APPEND CMAKE_CXX_FLAGS "-std=c++11")
+    list(APPEND CMAKE_CXX_FLAGS " -std=c++11 ")
   else()
     message ("** No CMAKE C++11 check. If the build breaks, you're on your own.")
   endif()
 endif ()
 
[...]

PR #421, opened 2017-10-10
Make cmake to better construct lists of parameters

benrubson:
Should help solving #420.


CMakeLists.txt
```diff
@@ -108,7 +108,7 @@ endif()
 set(CMAKE_REQUIRED_FLAGS "-Wl,--as-needed")
 CHECK_CXX_SOURCE_COMPILES("int main() {return 0;}" LINKER_SUPPORTS_WLASNEEDED)
 if(LINKER_SUPPORTS_WLASNEEDED)
-        list(APPEND EXTRA_LINKER_FLAGS "-Wl,--as-needed")
+        SET(EXTRA_LINKER_FLAGS "-Wl,--as-needed")
```
vgough: should append, not replace.

benrubson: This is our own variable which is not set above ?
I tested the following :
`SET(EXTRA_LINKER_FLAGS "${EXTRA_LINKER_FLAGS} -Wl,--as-needed")`
But then CMake complains because EXTRA_LINKER_FLAGS begins with a space !

vgough: ok, that makes sense.  Thanks.

mouse07410:
But this commit send to miss the $EXTRA_LINKER_FLAGS part after the -W...

mouse07410:
@benrubson ?

benrubson:
Sorry I got the notification but did not see the message in this PR, I then thought you deleted it...
So, `EXTRA_LINKER_FLAGS` is set to `-Wl,--as-needed`, what's the issue here @mouse07410 ?

CMakeLists.txt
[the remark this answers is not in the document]
benrubson: `EXTRA_LINKER_FLAGS` is not set elsewhere, so no need to transform this line 👍

PR #424, opened 2017-10-11
Correct last compilation warnings in readpassphrase

benrubson:
Hi,

This PR corrects last compilation warnings regarding `readpassphrase.cpp`.
It comes back to the original file, simply adding a dummy test to avoid the `(void)write()` warnings.
Will then help to progress on #426.

```
/home/makerpm/rpmbuild/BUILD/encfs-master/encfs/readpassphrase.cpp:128:46: warning: ignoring return value of 'ssize_t write(int, const void*, size_t)', declared with attribute warn_unused_result [-Wunused-result]
   (void)write(output, prompt, strlen(prompt));
                                              ^
/home/makerpm/rpmbuild/BUILD/encfs-master/encfs/readpassphrase.cpp:149:33: warning: ignoring return value of 'ssize_t write(int, const void*, size_t)', declared with attribute warn_unused_result [-Wunused-result]
     (void)write(output, "\n", 1);
                                 ^
```

Thank you !

### DIFF
--- a/encfs/readpassphrase.cpp
+++ b/encfs/readpassphrase.cpp
@@ -126,29 +126,30 @@ restart:
   }
 
   if (write(output, prompt, strlen(prompt)) != -1) {
-    end = buf + bufsiz - 1;
-    for (p = buf; (nr = read(input, &ch, 1)) == 1 && ch != '\n' && ch != '\r';) {
-      if (p < end) {
-        if ((flags & RPP_SEVENBIT) != 0) {
-          ch &= 0x7f;
-        }
-        if (isalpha(ch) != 0) {
-          if ((flags & RPP_FORCELOWER) != 0) {
-            ch = tolower(ch);
-          }
-          if ((flags & RPP_FORCEUPPER) != 0) {
-            ch = toupper(ch);
-          }
-        }
-        *p++ = ch;
-      }
-    }
-    *p = '\0';
+    //dummy test to get rid of warn_unused_result compilation warning
   }
+  end = buf + bufsiz - 1;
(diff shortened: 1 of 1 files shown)


vgough:
I'd rather not maintain a custom version of readpassphrase.  Do you know if upstream versions of this code still produce warnings?

benrubson:
Yes, I had a look, it does.
Finally I just transformed :
`(void)write`
into :
`if(write) {}`
To avoid clang warning of non-checked returned value.

rfjakob:
Shouldn't we error out if the write fails?

benrubson:
It's the `write()` which outputs the password prompt, so I think this is why it is simply not checked, password can be get even if prompt is not shown 👍

benrubson:
If we look at the modifications finally made to this upstream file, they are rather tiny :

The 2 write calls :
```
(void)write(output, prompt, strlen(prompt));
```
Have simply been changed to :
```
if (write(output, prompt, strlen(prompt)) != -1) {
  //dummy test to get rid of warn_unused_result compilation warning
}
```

That's it 👍 
And allow us to get rid of warning messages.
It's worth it I think :)

vgough:
Looks to me like a few things have changed.  I've created a PR #425 to import a new version.  I don't see any compiler warnings when building.

benrubson:
At least if we could commit this @vgough to get rid of compilation warnings + clang warnings visible in master CI, I think it would be perfect 👍 
Many thanks 👍

benrubson:
Many thanks @vgough 👍

Issue #407, opened 2017-10-03
Bunch of compiler warnings during compilation of current master current master


pbiering:
encfs-master/vendor/github.com/muflihun/easyloggingpp/src/easylogging++.h:
   1421x   defined but not used [-Wunused-variable]

others have marginal amount:

```
encfs-master/encfs/BlockFileIO.cpp:62:24: warning: comparison between signed and unsigned integer expressions [-Wsign-compare]
encfs-master/encfs/BlockFileIO.cpp:92:22: warning: comparison between signed and unsigned integer expressions [-Wsign-compare]
encfs-master/encfs/BlockFileIO.cpp:129:44: warning: comparison between signed and unsigned integer expressions [-Wsign-compare]
encfs-master/encfs/BlockFileIO.cpp:168:22: warning: comparison between signed and unsigned integer expressions [-Wsign-compare]
encfs-master/encfs/BlockFileIO.cpp:229:44: warning: comparison between signed and unsigned integer expressions [-Wsign-compare]
encfs-master/encfs/BlockFileIO.cpp:231:24: warning: comparison between signed and unsigned integer expressions [-Wsign-compare]
encfs-master/encfs/BlockFileIO.cpp:286:47: warning: comparison between signed and unsigned integer expressions [-Wsign-compare]
encfs-master/encfs/CipherFileIO.cpp:391:22: warning: comparison between signed and unsigned integer expressions [-Wsign-compare]
encfs-master/encfs/CipherFileIO.cpp:538:23: warning: comparison between signed and unsigned integer expressions [-Wsign-compare]
encfs-master/encfs/CipherFileIO.cpp:549:28: warning: comparison between signed and unsigned integer expressions [-Wsign-compare]

// solved (Ben) :
encfs-master/encfs/readpassphrase.cpp:128:46: warning: ignoring return value of 'ssize_t write(int, const void*, size_t)', declared with attribute warn_unused_result [-Wunused-result]
encfs-master/encfs/readpassphrase.cpp:149:33: warning: ignoring return value of 'ssize_t write(int, const void*, size_t)', declared with attribute warn_unused_result [-Wunused-result]
encfs-master/encfs/DirNode.cpp:524:21: warning: ignoring return value of 'int setfsuid(__uid_t)', declared with attribute warn_unused_result [-Wunused-result]
[...]

PR #425, opened 2017-10-13
update readpassphrase

vgough:
@benrubson This updates readpassphrase to the latest version I've found in a quick search.  I made a couple changes over the original:

* build as C++
* adjust include, add path.h include
* change NULL -> nullptr
* reformat with clang-format

### DIFF
--- a/encfs/readpassphrase.cpp
+++ b/encfs/readpassphrase.cpp
@@ -1,59 +1,47 @@
-/*	$OpenBSD: readpassphrase.c,v 1.12 2001/12/15 05:41:00 millert Exp $
+/*	$OpenBSD: readpassphrase.c,v 1.26 2016/10/18 12:47:18 millert Exp $
  */
 
 /*
- * Copyright (c) 2000 Todd C. Miller <Todd.Miller@courtesan.com>
- * All rights reserved.
+ * Copyright (c) 2000-2002, 2007, 2010
+ *	Todd C. Miller <Todd.Miller@courtesan.com>
  *
- * Redistribution and use in source and binary forms, with or without
- * modification, are permitted provided that the following conditions
- * are met:
- * 1. Redistributions of source code must retain the above copyright
- *    notice, this list of conditions and the following disclaimer.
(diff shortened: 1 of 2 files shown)


benrubson:
Clang-tidy complains a lot regarding this new version :)
And I think it should also complain (when some warnings will have been corrected ?) about the 2 `(void)write(...)`.

vgough:
There are still lots of clang-tidy complaints throughout encfs.   Looks like most of the complaints in readpassphrase are readability-implicit-bool-cast and readability-braces-around-statements, neither of which bother me much for imported C code.

I don't see any clang-tidy or compiler complaints about `(void)write`, so not sure what you're expecting.

benrubson:
Sorry, these `(void)write` warnings are not triggered by clang but by gcc... My mistake...
Taken from log given in #407 :
```
/home/makerpm/rpmbuild/BUILD/encfs-master/encfs/readpassphrase.cpp:128:46: warning: ignoring return value of 'ssize_t write(int, const void*, size_t)', declared with attribute warn_unused_result [-Wunused-result]
   (void)write(output, prompt, strlen(prompt));
                                              ^
/home/makerpm/rpmbuild/BUILD/encfs-master/encfs/readpassphrase.cpp:149:33: warning: ignoring return value of 'ssize_t write(int, const void*, size_t)', declared with attribute warn_unused_result [-Wunused-result]
     (void)write(output, "\n", 1);
                                 ^
```

I then would like to correct this, so that these warnings do not bother users anymore.
Thus I opened PR #424 (which I just updated with the log above to clarify).

For sure clang-tidy complaints regarding the new readpassphrase version are not so important, but I think it would be interesting to have clang-tidy to warn as less as possible, to easily detect future code mistakes. This is why I opened #427 (as well as #426), to try to make an effort to correct clang-tidy & compilation warnings.

Many thanks 👍

samrocketman:
@vgough this PR would be more appropriate as two commits because you're copying code.  1 commit for the original and 1 commit for your explicit changes.  This way your changes are more obviously documented because I can't tell what is your changes and what is the original copy at first glance.

Issue #427, opened 2017-10-14
Solve clang-tidy warnings


benrubson:
See [Travis CI log](https://travis-ci.org/vgough/encfs/branches) on master branch, the clang build, to list clang-tidy warnings to correct.

benrubson:
The following :
```
autosprintf.cpp:38:14: warning: do not define a C-style variadic function; consider using a function parameter pack or currying instead [cert-dcl50-cpp]
autosprintf::autosprintf(const char *format, ...) {
 ^
```
Leads to :
https://www.securecoding.cert.org/confluence/display/cplusplus/DCL50-CPP.+Do+not+define+a+C-style+variadic+function

And I then found several implementations :
https://gist.github.com/sehe/3374327
https://github.com/tromey/typesafe-printf
https://github.com/fmtlib/fmt

@vgough what do you think about this ?
Would you be OK to implement a solution to fix this warning ?
(I tried but I'm stuck somewhere trying to match the old constructor with the first method above...)
Many thx 👍

benrubson:
May I ask some help to correctly solve the following warnings ?
@rfjakob, would you please handle them ?
Will make the build log even more clean.
Thank you very much 👍 
```
/home/travis/build/vgough/encfs/encfs/DirNode.cpp:133:7: warning: class 'RenameOp' defines a destructor and a copy constructor but does not define a copy assignment operator, a move constructor or a move assignment operator [cppcoreguidelines-special-member-functions]
class RenameOp {
      ^
/home/travis/build/vgough/encfs/encfs/NullCipher.cpp:50:7: warning: class 'NullKey' defines a destructor but does not define a copy constructor, a copy assignment operator, a move constructor or a move assignment operator [cppcoreguidelines-special-member-functions]
class NullKey : public AbstractCipherKey {
      ^
/home/travis/build/vgough/encfs/encfs/NullCipher.cpp:56:7: warning: class 'NullDestructor' defines a destructor, a copy constructor and a copy assignment operator but does not define a move constructor or a move assignment operator [cppcoreguidelines-special-member-functions]
class NullDestructor {
      ^
/home/travis/build/vgough/encfs/encfs/SSL_Cipher.cpp:277:7: warning: class 'SSLKey' defines a destructor but does not define a copy constructor, a copy assignment operator, a move constructor or a move assignment operator [cppcoreguidelines-special-member-functions]
class SSLKey : public AbstractCipherKey {
      ^
[...]

PR #429, opened 2017-10-20
Correct ato* clang warnings

benrubson:
Hi,

This PR helps with #427, it solves clang warnings regarding ato* functions.

Thank you 👍 

Ben

### DIFF
--- a/encfs/FileUtils.cpp
+++ b/encfs/FileUtils.cpp
@@ -644,7 +644,7 @@ static Cipher::CipherAlgorithm selectCipherAlgorithm() {
     cout << "\n" << _("Enter the number corresponding to your choice: ");
     char answer[10];
     char *res = fgets(answer, sizeof(answer), stdin);
-    int cipherNum = (res == nullptr ? 0 : atoi(answer));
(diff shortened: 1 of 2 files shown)


encfs/FileUtils.cpp
```diff
@@ -644,7 +644,7 @@ static Cipher::CipherAlgorithm selectCipherAlgorithm() {
     cout << "\n" << _("Enter the number corresponding to your choice: ");
     char answer[10];
     char *res = fgets(answer, sizeof(answer), stdin);
-    int cipherNum = (res == nullptr ? 0 : atoi(answer));
+    int cipherNum = (res == nullptr ? 0 : (int)strtol(answer, nullptr, 10));
```
rfjakob: Hmm, where is the improvement here? Still no check if we got a valid number.

benrubson: Thx for your review !
Goal is at least to remove warnings.
There is a check just after `strtol()`, so it's OK.
In `XmlReader.cpp`, I've added checks which were not done before.
Thx !

Issue #426, opened 2017-10-14
Solve compilation warnings


benrubson:
See [Travis CI log](https://travis-ci.org/vgough/encfs/branches) on master branch, the gcc build, to list compilation warnings to correct.

benrubson:
OK, one type of warnings remaining, it's regarding comparisons between signed and unsigned (-Wsign-compare).

Not sure how to handle this properly.
Looking at the code, perhaps we should begin moving blockSize from int to unsigned int.
And carefully check other operations based on blockSize.
(as for example an unsigned int can't fit into an int)

Any thoughts ?

benrubson:
OK, last compilation warning solved 👍

PR #432, opened 2017-10-20
Correct unused-parameter warnings

benrubson:
Hi,

This PR helps with #427 and #426, it solves warnings regarding unused-parameter.

Thank you 👍 

Ben

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -180,7 +180,6 @@ if (${CMAKE_MAJOR_VERSION}.${CMAKE_MINOR_VERSION} GREATER 3.5) # Need 3.6 or abo
       ",-google-runtime-int"
       ",-google-runtime-references"
       ",-misc-misplaced-widening-cast"
-      ",-misc-unused-parameters"
       ",-modernize-loop-convert"
       ",-readability-inconsistent-declaration-parameter-name"
       ",-readability-named-parameter"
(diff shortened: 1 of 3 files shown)


CMakeLists.txt
```diff
@@ -180,7 +180,6 @@ if (${CMAKE_MAJOR_VERSION}.${CMAKE_MINOR_VERSION} GREATER 3.5) # Need 3.6 or abo
       ",-google-runtime-int"
       ",-google-runtime-references"
       ",-misc-misplaced-widening-cast"
-      ",-misc-unused-parameters"
```
rfjakob: Why remove the check?

encfs/main.cpp
```diff
@@ -621,7 +619,7 @@ int main(int argc, char *argv[]) {
   // encfs_oper.releasedir = encfs_releasedir;
   // encfs_oper.fsyncdir = encfs_fsyncdir;
   encfs_oper.init = encfs_init;
-  encfs_oper.destroy = encfs_destroy;
+  // encfs_oper.destroy = encfs_destroy;
```
rfjakob: Unsure about this. Leaving encfs_oper.destroy = NULL does not change the behavoir ob libfuse?

In any case, should be deleted instead of commented out.

benrubson: I verified this, and no, the function is simply called if it exists, but is not replaced by anything else : https://github.com/libfuse/libfuse/blob/fuse_2_9_5/lib/fuse.c#L2614
I just removed the comment.
Thx 👍

CMakeLists.txt
benrubson: All checks are enabled, but the listed ones, so I simply re-enable `misc-unused-parameters` 👍

benrubson: OK for you @rfjakob ? Thx 👍

PR #433, opened 2017-10-20
Correct unnecessary-value-param clang warnings

benrubson:
Hi,

This PR helps with #427, it solves clang warnings regarding unnecessary-value-param.
**Please review with care.**

Thx 👍 

Ben


encfs/base64.cpp
```diff
@@ -252,7 +252,7 @@ bool B64StandardDecode(unsigned char *out, const unsigned char *in, int inLen) {
 // If you want to use an alternate alphabet, change the characters here
 const static char encodeLookup[] =
     "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789+/";
-std::string B64StandardEncode(std::vector<unsigned char> inputBuffer) {
+std::string B64StandardEncode(std::vector<unsigned char> &inputBuffer) {
```
rfjakob: Shouldn't inputBuffer be const?

benrubson: Right. Just committed 👍

PR #436, opened 2017-10-22
Add a lock condition before idle unmount

benrubson:
Hi,

As documented into the updated man page, this PR adds a lock condition when using `--idle` :
_EncFS will try to lock mount point just before unmounting, and will unmount only if lock succeeds, giving users the ability to handle races when EncFS is used in automated tools or scripts._

This allows for example an automated tool to lock the mount point before checking if it is already mounted, and mount it if needed, avoiding races with `--idle` if mount point is already mounted.

Thank you very much 👍 

Ben

### DIFF
--- a/encfs/main.cpp
+++ b/encfs/main.cpp
@@ -784,8 +784,8 @@ static void *idleMonitor(void *_arg) {
         RLOG(WARNING) << "Filesystem inactive, but " << openCount
                       << " files opened: " << arg->opts->mountPoint;
       } else {
-      	/* try to lock the mount point to give an oportunity to handle races
-      	   with an automated tool or script using --idle. */
+        /* try to lock the mount point to give an oportunity to handle races
+           with an automated tool or script using --idle. */
         int fdlock = open(arg->opts->mountPoint.c_str(),O_RDONLY);
         if (flock(fdlock, LOCK_EX|LOCK_NB) == -1) {
           RLOG(WARNING) << "Filesystem inactive, but "
--- a/encfs/main.cpp
+++ b/encfs/main.cpp
@@ -22,6 +22,7 @@
 #include <cstring>
 #include <ctime>
 #include <exception>
+#include <sys/file.h>
 #include <getopt.h>
 #include <iostream>
 #include <memory>
@@ -765,7 +766,7 @@ static void *idleMonitor(void *_arg) {
   pthread_mutex_lock(&ctx->wakeupMutex);
 
   while (ctx->running) {
-    int usage, openCount;
+    int usage, openCount, fdlock;
     ctx->getAndResetUsageCounter(&usage, &openCount);
 
     if (usage == 0 && ctx->isMounted()) {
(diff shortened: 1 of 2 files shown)


encfs/main.cpp
```diff
@@ -779,16 +780,21 @@ static void *idleMonitor(void *_arg) {
     }
 
     if (idleCycles >= timeoutCycles) {
-      if (openCount == 0) {
+      if (openCount != 0) {
+        RLOG(WARNING) << "Filesystem inactive, but " << openCount
+                      << " files opened: " << arg->opts->mountPoint;
+      } else if ((fdlock = open(arg->opts->mountPoint.c_str(),O_RDONLY)) == -1 || flock(fdlock, LOCK_EX|LOCK_NB) == -1) {
```
rfjakob: Way too much stuff if that if(), IMO

rfjakob:
Could you explain again who would use that function, and how?

encfs/main.cpp
benrubson: I rewrited it 👍

benrubson:
> Could you explain again who would use that function, and how?

Sure.
Let's imagine an automated tool which mounts EncFS FS, based for example on users activity. or a service, or whatever...
This tool mounts with `--idle`, so that it is sure FS will be unmounted once inactive, without having to ask users to do it. Perfect.

Now, let's imagine one FS is mounted, and very closed to the inactivity timeout.
And let's imagine that the automated tool needs to verify now that the FS is mounted, and if it is not, to mount it.
Here there is a race.

The maintainer of this automated tool can then enclose the check & mount operations in a lock of the mount point, to be sure there will be no race :
```
lock_mount_point
check_if_already_mounted
mount_if_needed
touch_mount_point_to_reset_idle_counter
unlock_mount_point
```
Here, EncFS will not be able to lock the mount point, and then will not unmount 👍

rfjakob:
How about
```
touch_mount_point_to_reset_idle_counter
check_if_already_mounted
mount_if_needed
```
?

benrubson:
There is still a race if the 2 first operations are done while EncFS is unmouting.

rfjakob:
I guess you mean the between 

https://github.com/vgough/encfs/blob/92b209f7c021fe997893dafe82a6b6049bf3c1af/encfs/main.cpp#L766
and 
https://github.com/vgough/encfs/blob/92b209f7c021fe997893dafe82a6b6049bf3c1af/encfs/main.cpp#L780 .

That's true, there is a race. But this race is internal to EncFS, and can and should be handled internally.

benrubson:
Not necessarily, I also mean while fuse is unmounting the FS.
If there is no lock here to avoid actions to be done concurrently, FS may appear as mounted to the external tool, while it is unmouting.

rfjakob:
Yes we need a lock to prevent "touch_mount_point_to_reset_idle_counter" from returning without resetting the idle counter

benrubson:
What EncFS would do if it receives IO operations while it is locking to unmount ?
Block the operations I assume ?
Would it return something to this operations after the unmount ?

rfjakob:
Block or return an error immediately

benrubson:
I have a working fix for this, I plan to submit it shortly 👍

benrubson:
Please see #448 for another solution.
Thx 👍

PR #444, opened 2017-11-02
Correct readability-non-const-parameter warning

benrubson:
Hi,

This helps solving #427, correcting the following :
```
SSL_Cipher.cpp:508:59: warning: pointer parameter 'chainedIV' can be pointer to const [readability-non-const-parameter]
                             int dataLen, uint64_t *const chainedIV) {
                                                          ^
                                          const 
```

Ben

### DIFF
--- a/encfs/SSL_Cipher.cpp
+++ b/encfs/SSL_Cipher.cpp
@@ -505,7 +505,7 @@ CipherKey SSL_Cipher::newRandomKey() {
     compute a 64-bit check value for the data using HMAC.
 */
 static uint64_t _checksum_64(SSLKey *key, const unsigned char *data,
-                             int dataLen, const uint64_t *chainedIV) {
+                             int dataLen, const uint64_t *const chainedIV) {
   rAssert(dataLen > 0);
   Lock lock(key->mutex);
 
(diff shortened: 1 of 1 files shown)


encfs/SSL_Cipher.cpp
```diff
@@ -505,7 +505,7 @@ CipherKey SSL_Cipher::newRandomKey() {
     compute a 64-bit check value for the data using HMAC.
 */
 static uint64_t _checksum_64(SSLKey *key, const unsigned char *data,
-                             int dataLen, uint64_t *const chainedIV) {
+                             int dataLen, const uint64_t *chainedIV) {
```
rfjakob: const uint64_t *const chainedIV ?

benrubson: Any reason why we need a `*const` here ?

benrubson: Just as a security, to be sure not to accidentally modify chainedIV pointer later in the function's code ?
Which is not the case actually.
Thx 👍

rfjakob: Yes. More const is more better, right?

rfjakob: Your commit does two things:
1) It makes one variable const (the variable chainedIV points to)
2) Makes another non-const (the pointer itself).
I agree with (1) but I don't understand (2).

PR #445, opened 2017-11-02
Correct performance warnings

benrubson:
Hi,

This PR helps solving #427, correcting :

```
DirNode.cpp:423:41: warning: string concatenation results in allocation of unnecessary temporary strings; consider using 'operator+=' or 'string::append()' instead [performance-inefficient-string-concatenation]
      string newFull = sourcePath + '/' + newName;
                                        ^
```
_here it disables the Clang test_

However, I'm not sure how to handle this one :
```
FileUtils.cpp:500:58: warning: the parameter 'data' is copied for each invocation but only used as a const reference; consider making it a const reference [performance-unnecessary-value-param]
                              std::vector<unsigned char> data) {
                                                         ^
```

Edit : just found a way, please **review with care** !

Thx 👍 

Ben

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -180,6 +180,7 @@ if (${CMAKE_MAJOR_VERSION}.${CMAKE_MINOR_VERSION} GREATER 3.5) # Need 3.6 or abo
       ",-google-runtime-int"
       ",-google-runtime-references"
       ",-modernize-loop-convert"
+      ",-performance-inefficient-string-concatenation"
       ",-readability-inconsistent-declaration-parameter-name"
       ",-readability-named-parameter"
       )
--- a/encfs/FileUtils.cpp
+++ b/encfs/FileUtils.cpp
@@ -465,7 +465,7 @@ bool saveConfig(ConfigType type, const string &rootDir,
 template <typename T>
 tinyxml2::XMLElement *addEl(tinyxml2::XMLDocument &doc,
                             tinyxml2::XMLNode *parent, const char *name,
-                            T value) {
+                            const T &value) {
   auto el = doc.NewElement(name);
   el->SetText(value);
   parent->InsertEndChild(el);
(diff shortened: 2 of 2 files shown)


rfjakob:
Unsure about these changes. Looks like we lose a lot of readability.

Compare

       string plainPath = volumeDir + name;

With

       string plainPath = string(volumeDir).append(name);

benrubson:
Goal was to conform to clang rules to get rid of the warnings.
We also get rid of some temporary buffers using append().
We could try to use += to increase readability ? However I'm not sure it would really help.
Lines of code are a little bit longer, but still readable IMO, and almost limited to one file (DirNode.cpp).

benrubson:
What about the second commit ?
OK with it @rfjakob ?
Thx 👍

rfjakob:
"Correct unnecessary-value-param" looks good to me

benrubson:
Thx for your review 👍 
So what about this whole PR ?
Would you be ready to make a small concession on readability (in almost only one file) to correct warnings ?

rfjakob:
To be honest, I see the downside (more complicated code), but I don't see the upside yet. How many % performance do we gain?

benrubson:
Of course hard to measure, almost nothing I would say.
Primary goal is to get ride of all compilation & clang warnings, so that we have a clear build log, making it easier to check if there are new warnings when a PR / new code is submitted / merged.

rfjakob:
I think the situation would be different if this was a correctness problem. But as this is only a performance warning, I suggest to turn the warning off.

rfjakob:
(I agree with your goal of getting the code warning-free!)

benrubson:
I like when we disable as least tests as possible, but OK, let's then disable `performance-inefficient-string-concatenation`.

benrubson:
OK, PR updated 👍

PR #446, opened 2017-11-02
Correct sign-compare warnings

benrubson:
Hi,

This PR helps solving #426.
It adresses sign-compare warnings.

Ben


rfjakob:
Rather than all these casts, shouldn't the variables / return values be off_t or size_t ?

benrubson:
Read and write operations return ssize_t, as they can return values < 0 in case of error.
This is I think what implies the casts.
Perhaps you think about something else ?

encfs/BlockFileIO.cpp
```diff
@@ -146,7 +146,7 @@ ssize_t BlockFileIO::read(const IORequest &req) const {
 
     // if we're reading a full block, then read directly into the
     // result buffer instead of using a temporary
-    if (partialOffset == 0 && size >= (size_t)_blockSize) {
+    if (partialOffset == 0 && size >= _blockSize) {
```
rfjakob: Shouldn't _blockSize be unsigned from the start?

encfs/BlockFileIO.cpp
```diff
@@ -164,8 +164,8 @@ ssize_t BlockFileIO::read(const IORequest &req) const {
       break;  // didn't get enough bytes
     }
 
-    size_t cpySize = min((size_t)(readSize - partialOffset), size);
-    CHECK(cpySize <= readSize);
+    size_t cpySize = min((size_t)readSize - (size_t)partialOffset, size);
```
rfjakob: Shouldn't partialOffset be unsigned?

encfs/BlockFileIO.cpp
```diff
@@ -256,7 +256,7 @@ ssize_t BlockFileIO::write(const IORequest &req) {
     // if writing an entire block, or writing a partial block that requires
     // no merging with existing data..
     if ((toCopy == _blockSize) ||
-        (partialOffset == 0 && blockReq.offset + toCopy >= fileSize)) {
+        (partialOffset == 0 && blockReq.offset + (off_t)toCopy >= fileSize)) {
```
rfjakob: toCopy is not signed, is it?

encfs/BlockFileIO.cpp
benrubson: It is unsigned now, thanks to this PR.
Here is then remove a wast which is no more needed, as size_t is unsigned.

encfs/CipherFileIO.cpp
```diff
@@ -535,7 +535,7 @@ ssize_t CipherFileIO::read(const IORequest &origReq) const {
    * to the data. */
   if (req.offset < 0) {
     headerBytes = -req.offset;
-    if (req.dataLen < headerBytes) {
+    if (req.dataLen < (size_t)headerBytes) {
```
rfjakob: Shouldn't headerBytes be unsigned?

encfs/BlockFileIO.cpp
benrubson: No, so without the cast this leads to a comparison between size_t and off_t :
`comparison of integers of different signs: 'unsigned long' and 'off_t' (aka 'long') [clang-diagnostic-sign-compare]`

encfs/BlockFileIO.cpp
benrubson: Literally yes, but as req.offset is signed, I would say no.
In addition, AFAIR, min() expect 2 members of the same type, so cast would be required.

encfs/CipherFileIO.cpp
benrubson: Yep, but HEADER_SIZE itself is a int...
Moving headerBytes to unsigned int would then lead to some limit considerations, for example line 545.
You point to #447 :)

benrubson: I had an additional look, if we make HEADER_SIZE unsigned, other casts will be needed as it is compared to st_size which is an off_t...

rfjakob: Hmm. Maybe we should move everything to ssize_t or int64_t.

benrubson: I also thought about it, but I think `dataLen` should remain `size_t`, as in `read(2)` and `write(2)`.

rfjakob:
I'll have to think more about this. My problem with the casts is that they don't make the code safer.

What do you run to get the warnings? Clang-tidy?

benrubson:
Yep, Clang-tidy.
[Travis runs it](https://travis-ci.org/vgough/encfs/branches), look at the Clang C++ job.

Going through the code, there are many places where size_t / ssize_t / off_t / int... are wrongly used and may lead to some overflows... (this is why I opened #447)
This is a quite huge check to perform.
Look for example here, where an int is used to go through the number of blocks (should be an off_t) :
https://github.com/vgough/encfs/blob/v1.9.2/encfs/encfsctl.cpp#L388

I think we should keep :
- size_t for dataLen, as in read(2) / write(2) ;
- ssize_t as read & write return type, as in read(2) / write(2) ;
- off_t for sizes and offsets, as in stat(2).

I did a huge code review a few weeks ago, but taking care of functions return code values, especially to get correct RC from IO operations. I did not perform a full types review.

encfs/CipherFileIO.cpp
```diff
@@ -546,7 +546,7 @@ ssize_t CipherFileIO::read(const IORequest &origReq) const {
     memcpy(req.data, &headerBuf[headerOffset], headerBytes);
 
     // the read does not want data beyond the header
-    if (headerBytes == req.dataLen) {
+    if ((size_t)headerBytes == req.dataLen) {
```
rfjakob: Ok I looked through all of the casts again. Everything seems safe except this one. It looks like headerBytes may be negative at this point. Rather cast req.dataLen to int?

benrubson: Mmmmh I'm not sure how `headerBytes` could be negative.
Which path would make it negative ?

I think casting `size_t` to `int` is rather dangerous.

Thx 👍

benrubson:
Thank you for your review @rfjakob.
So we will go through whole types and casts, as we thought about above, later / in another PR ?
Thx 👍

PR #448, opened 2017-11-06
Remove race in idle unmount

### LINKED ISSUE #344: Race conditions in idle checker
I think there are a couple of race conditions in the idle monitor. This is based on reviewing the code, they seem hard to trigger so it is not a show stopper.

1) in main.cpp before the `idleMonitor` there is a comment saying that we check if there are any open files before unmounting the file system, but there is no mutex that stops another thread from opening a new file between the open file count check and unmounting the filesystem.

2) In `EncFS_Context::isMounted` is not protected by any mutex.

3) `EncFS_Context::getRoot` will remount the file system when needed, but doesn't check if some other thread already is performing the remount.

benrubson:
Hi,

This PR solves the race that exists in idle unmount, discussed in #436.
This then closes #344, and of course also closes #436.

2 solutions proposed through 2 commits, but I think, as explained in my inline comment below, that v1 should contain an issue.
The other solution, v2, is even more simple.

Thank you 👍 

Ben

### DIFF
--- a/encfs/Context.cpp
+++ b/encfs/Context.cpp
@@ -32,15 +32,14 @@ EncFS_Context::EncFS_Context() {
   pthread_cond_init(&wakeupCond, nullptr);
   pthread_mutex_init(&wakeupMutex, nullptr);
   pthread_mutex_init(&contextMutex, nullptr);
-  pthread_mutex_init(&unmountMutex, nullptr);
 
   usageCount = 0;
   idleCount = -1;
+  isUnmounting = false;
   currentFuseFh = 1;
 }
 
 EncFS_Context::~EncFS_Context() {
-  pthread_mutex_destroy(&unmountMutex);
   pthread_mutex_destroy(&contextMutex);
   pthread_mutex_destroy(&wakeupMutex);
   pthread_cond_destroy(&wakeupCond);
@@ -52,12 +51,11 @@ std::shared_ptr<DirNode> EncFS_Context::getRoot(int *errCode) {
   std::shared_ptr<DirNode> ret = nullptr;
   do {
     {
-      Lock locku(unmountMutex, false);
-      if(!locku.isLocked()) {
-      	*errCode = EBUSY;
-      	break;
-      }
       Lock lock(contextMutex);
+      if (isUnmounting) {
+        *errCode = -EBUSY;
+        break;      	
+      }
       ret = root;
       ++usageCount;
     }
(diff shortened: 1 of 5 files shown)


encfs/Context.cpp
```diff
@@ -46,9 +49,14 @@ EncFS_Context::~EncFS_Context() {
   openFiles.clear();
 }
 std::shared_ptr<DirNode> EncFS_Context::getRoot(int *errCode) {
-  std::shared_ptr<DirNode> ret;
+  std::shared_ptr<DirNode> ret = nullptr;
   do {
     {
+      Lock locku(unmountMutex, false);
+      if(!locku.isLocked()) {
```
benrubson: Mmmmh I think there might be an issue here if 2 IO operations call `getRoot()` at the same time, one of the 2 may be dropped...
This non-bloking lock allows `unmount()` to proceed correctly, but I finally think there is a drawback...

benrubson: Other solution found, I'll submit it

benrubson:
Your thoughts on this @rfjakob perhaps ?
Many thx 👍

rfjakob:
I like the diffstat. Will review in detail tomorrow!

encfs/Context.cpp
```diff
@@ -75,15 +81,39 @@ void EncFS_Context::setRoot(const std::shared_ptr<DirNode> &r) {
   }
 }
 
-bool EncFS_Context::isMounted() { return root != nullptr; }
-
-void EncFS_Context::getAndResetUsageCounter(int *usage, int *openCount) {
+bool EncFS_Context::usageAndUnmount(int timeoutCycles) {
```
rfjakob: This function needs a comment on what it does and what the return code means

encfs/Context.cpp
```diff
@@ -75,15 +81,39 @@ void EncFS_Context::setRoot(const std::shared_ptr<DirNode> &r) {
...
+    }
+    else {
+      idleCount = 0;
+    }
+    VLOG(1) << "idle cycle count: " << idleCount << ", timeout at "
+            << timeoutCycles;
+
+    usageCount = 0;
+
+    if (idleCount >= timeoutCycles) {
```
rfjakob: if (idleCount < timeoutCycles) {
    return false;
    }

Allows you to get rid of one indentation level

encfs/Context.cpp
```diff
@@ -75,15 +81,39 @@ void EncFS_Context::setRoot(const std::shared_ptr<DirNode> &r) {
...
+      if (!openFiles.empty()) {
+        if (idleCount % timeoutCycles == 0) {
+          RLOG(WARNING) << "Filesystem inactive, but " << openFiles.size()
+                        << " files opened: " << this->opts->mountPoint;
+        }
```
rfjakob: "return false" to get rid of the else. The if chain is three levels deep and hard to follow

encfs/Context.cpp
benrubson: done

encfs/Context.cpp
benrubson: and done !
Thx for your review 👍

benrubson:
I think I need you to approve @rfjakob so that I can merge, if it's OK for you.
Thank you 👍

benrubson:
Perfect thank you 👍

PR #463, opened 2018-01-28
Fix encfs build with BUILD_SHARED_LIBS=1

vaygr:
This fixes 2 things when building encfs with `BUILD_SHARED_LIBS=1`:

- tinyxml2 target is different when it's built as a shared library, resulting into `tinyxml2.so` for linking via `-ltinyxml2`
- easyloggingpp is a header library, so it [doesn't have](https://github.com/muflihun/easyloggingpp/issues/603) an option to build as shared, and turning off static simply meant _no_ library building at all


benrubson:
Many thx @vaygr 👍 

Mmmh I'm afraid we will certainly forget and loose your modifications in `vendor/` path when we will update the library ?

Perhaps we should build/force easyloggingpp as static until a shared option is added by the author to the mainstream (we would then simply have to modify our `CMakeLists.txt` file) ?

A `BUILD_SHARED_LIBS` Travis job could also help us not to forget :)

vaygr:
Yeah, it doesn't take `git-vendor` usage into account. So we could wait until that upstream issue is closed.

vaygr:
I split the commits for tinyxml2/easyloggingpp, and reverted the last for historical purposes.

This will force building easyloggingpp as static till the source issue is fixed by upstream.

benrubson:
I think this is the way to go, thank you 👍 
I'll let @rfjakob do an additional review and merge the patch.

benrubson:
Mmmh Travis is stuck dealing with cmake.org :)

PR #465, opened 2018-01-29
Correct FSF address.

Vascom:
Corrected FSF address in source files corresponding to https://www.gnu.org/licenses/old-licenses/gpl-2.0.txt

### DIFF
--- a/encfs/autosprintf.cpp
+++ b/encfs/autosprintf.cpp
@@ -14,8 +14,8 @@
 
    You should have received a copy of the GNU Library General Public
    License along with this program; if not, write to the Free Software
-   Foundation, Inc., 59 Temple Place - Suite 330, Boston, MA 02111-1307,
-   USA.  */
+   Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301
+   USA. */
 
 /* Tell glibc's <stdio.h> to provide a prototype for vasprintf().
    This must come before <config.h> because <config.h> may include
--- a/encfs/autosprintf.cpp
+++ b/encfs/autosprintf.cpp
@@ -2,20 +2,18 @@
    Copyright (C) 2002 Free Software Foundation, Inc.
    Written by Bruno Haible <bruno@clisp.org>, 2002.
 
-   This program is free software; you can redistribute it and/or modify it
-   under the terms of the GNU Library General Public License as published
-   by the Free Software Foundation; either version 2, or (at your option)
-   any later version.
+   This program is free software: you can redistribute it and/or modify
+   it under the terms of the GNU Library General Public License as published by
+   the Free Software Foundation; either version 2 of the License, or
+   (at your option) any later version.
 
    This program is distributed in the hope that it will be useful,
    but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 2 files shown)


Vascom:
May be change it to addressless license notice https://www.gnu.org/licenses/gpl-howto.en.html ?

rfjakob:
Yes, please go with this ( from https://www.gnu.org/licenses/gpl-howto.en.html ):
```
    You should have received a copy of the GNU General Public License
    along with this program.  If not, see <http://www.gnu.org/licenses/>.
```

encfs/autosprintf.cpp
```diff
@@ -2,20 +2,18 @@
...
    Written by Bruno Haible <bruno@clisp.org>, 2002.
 
-   This program is free software; you can redistribute it and/or modify it
-   under the terms of the GNU Library General Public License as published
-   by the Free Software Foundation; either version 2, or (at your option)
-   any later version.
+   This program is free software: you can redistribute it and/or modify
+   it under the terms of the GNU Library General Public License as published by
+   the Free Software Foundation; either version 2 of the License, or
+   (at your option) any later version.
```
samrocketman: Note: this comment does not block merging.

This phrasing is already in here (I didn't notice it before).  Something to consider going forward: https://opensource.stackexchange.com/questions/1777/why-is-gplv2-incompatible-with-gplv3

Quote from commenter "DevSolar":

> GPLv2+ code that's been "upped" to GPLv3 has to remain GPLv3. The person you got the GPLv2+ code from has no way to merge back any downstream (GPLv3) changes back into his (GPLv2+) codebase.

Should we consider removing `or (at your option) any later version` or accept if anybody just so happens to want to merge GPLv3 code back to us we decide later to upgrade this project to GPLv3?

In general, I'm not personally comfortable with open ended legal comments like that.

Vascom: It is normal ended and the same phrase was in original license notice.

benrubson:
Is this ready to be merged ?

samrocketman:
I don't have any blocking comments.  I was just discussing.

benrubson:
OK 👍 I'm not sure about what is really needed here, so I'm just asking :)

samrocketman:
Ok well I guess to be clear I leave this up to the current project maintainers to decide.  I was just asking questions.  So I guess the decision falls to you @benrubson or if you're not comfortable merging ask another maintainer to decide to block or merge this change.

From a licensing perspective it's okay to merge in my opinion since it's not changing any terms, unlucky updating a point of contact.

rfjakob:
Looks good to me.

encfs/autosprintf.cpp
samrocketman: I don't think it's normal nor acceptable to have such open ended terms.  My personal opinion of course.  I would be OK if GPLv3 was compatible with GPLv2 but that's not how it is in the currently language of GPLv3.  "Or later version" is super open ended legally and could mean anything down the line.  I respectfully disagree with your opinion.

rfjakob: Not sure what you mean with "not normal", this is the standard text from GNU https://www.gnu.org/licenses/gpl-2.0.html#SEC4

samrocketman: In legal terms it's not normal in my opinion.

PR #471, opened 2018-02-23
Do not count usage on root path

benrubson:
Hi,

On some systems, at least on FreeBSD, an user which is not allowed to access an EncFS folder can at least stat its mountpoint.
This user can then prevent the `--idle` option to work regularly stating the mountpoint.

This PR corrects this issue dropping usage counts made on root path `/`.

Thank you 👍 

Ben

### DIFF
--- a/encfs/Context.cpp
+++ b/encfs/Context.cpp
@@ -47,7 +47,12 @@ EncFS_Context::~EncFS_Context() {
   // release all entries from map
   openFiles.clear();
 }
+
 std::shared_ptr<DirNode> EncFS_Context::getRoot(int *errCode) {
+  return getRoot(errCode, false);
+}
+
+std::shared_ptr<DirNode> EncFS_Context::getRoot(int *errCode, bool skipUsageCount) {
   std::shared_ptr<DirNode> ret = nullptr;
   do {
     {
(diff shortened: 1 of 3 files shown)


rfjakob:
Just FYI, does not work on Linux:
```
[root@brikett tmp.CODdZ1YXtD]# ls -l
ls: cannot access 'b': Permission denied
d?????????? ? ?     ?      ?            ? b
```
Some more comments:
1) This needs an `#ifdef` or something to only activate it on freebsd
2) Using the output variable `errCode`  to pass input is hairy
3) What else beside `stat()` on the mountpoint will it ignore?

benrubson:
Sounds like you can't even `stat` the mount point on linux.
If you do, does the `stat` arrive to EncFS (do you see it in EncFS foreground verbose log) ?

Regarding your comments :
1. Yes if Linux is not concerned by this, why not.
2. I agree, but it was the less intrusive method I found. Let's for for an additional parameter to `getRoot()` ?
3. stat, touch... every single operation on the mount point itself will not reset the timeout. Well generally we do at least a directory listing of the mount point... In #436 we talked about `touch_mount_point_to_reset_idle_counter`, it would now then rather be `ls_mount_point_to_reset_idle_counter`...

rfjakob:
No, nothing arrives at EncFS. The kernel blocks completely.

Yes I think adding a second parameter to `getRoot()` is better. Something like

    getRoot(int *errCode, bool skipUsageCount) 

You can define the one-parameter variant to call `getRoot(errCode, false)` so you don't have to change the callers.

benrubson:
Thx 👍

PR #476, opened 2018-03-10
Correct MAC typo in FS creation

benrubson:


mouse07410:
I consider this change trivial enough to merge quickly. @rfjakob would you agree?

PR #478, opened 2018-03-12
Ignore .encfs6.xml file in reverse mode

### LINKED ISSUE #101: encfsctl showcruft shows ~/.encfs6.xml
If I use `encfsctl showcruft` on an (unmounted) encfs-encrypted directory (which otherwise works fine):

```
encfsctl showcruft ~/Dropbox/.encfs-crypted
EncFS Password:
In directory /:
/Users/myusername/Dropbox/.encfs-crypted//.encfs6.xml
Found 1 invalid file.
```

As you can see, it identifies `.encfs6.xml` as cruft. But surely this is the very opposite? This file is essential to encfs' working. Granted, it technically isn't an encrypted file for the filesystem, but it would seem logical to exclude it from showcruft, since it will always be present.

I am using encfs 1.8.1 on OS X 10.10.4. I also see the same problem with encfs 1.7.4 on Ubuntu 14.04.2.

### LINKED ISSUE #174: .encfs6.xml treated as file when using no filename encryption
When using no (‘Null’) filename encryption, available through expert configuration mode, `.encfs6.xml` is displayed as a file in the mounted volume, and can even be deleted. Attempting to access the file gives gibberish, or causes an I/O error if MAC verification is enabled.

Vaguely related to #101 in that both relate to encfs being blind to the special nature of `.encfs6.xml`.

benrubson:
Hi,

This PR hides the configuration file `.encfs6.xml` in reverse mode.
It then closes #101 and closes #174.

Thank you for your review 👍 

Ben

### DIFF
--- a/encfs/DirNode.cpp
+++ b/encfs/DirNode.cpp
@@ -52,8 +52,8 @@ class DirDeleter {
 };
 
 DirTraverse::DirTraverse(std::shared_ptr<DIR> _dirPtr, uint64_t _iv,
-                         std::shared_ptr<NameIO> _naming)
-    : dir(std::move(_dirPtr)), iv(_iv), naming(std::move(_naming)) {}
+                         std::shared_ptr<NameIO> _naming, bool _root)
+    : dir(std::move(_dirPtr)), iv(_iv), naming(std::move(_naming)), root(_root) {}
 
 DirTraverse &DirTraverse::operator=(const DirTraverse &src) = default;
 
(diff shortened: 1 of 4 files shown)


rfjakob:
Good idea, gocryptfs does the same. But this needs a testcase to verify that it works as intendend, and keeps working as indented in the future

rfjakob:
We should test two things:

1) plain/.encfs6.xml should not show up in cipher
2) plain/foo/.encfs6.xml must show up in cipher

benrubson:
Thx for your review, test cases added 👍

PR #482, opened 2018-03-15
Revert from/to link parameters for readability

### LINKED ISSUE #475: encfs_link and DirNode::link methods parameters named wrong
In encfs.cpp, function encfs_link is defined this way:
int encfs_link(const char *from, const char *to)

And in DirNode.cpp:
int DirNode::link(const char *from, const char *to)

The parameters are named wrong. The first parameter should be "to" and the second "from".

This is not a bug, just causes confusion when analysing the code.
The parameters should be named in order equal to encfs_symlink, witch is correct.

benrubson:
Hi,

This is a cosmetic change which closes #475.

Ben

### DIFF
--- a/encfs/DirNode.cpp
+++ b/encfs/DirNode.cpp
@@ -609,14 +609,14 @@ int DirNode::rename(const char *fromPlaintext, const char *toPlaintext) {
   return res;
 }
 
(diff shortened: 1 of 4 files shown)


benrubson:
Thx for the approval @rfjakob 👍

Issue #447, opened 2017-11-02
Implicit casts should be checked


benrubson:
Here is a bug report in order to think about types' check.
There are some implicit casts between size_t / ssize_t / int which are sometimes wrongly done, leading to some limits (even if they are high) which could be overflowed.
For example, read and write operations deal with as much as dataLen (size_t), and return ssize_t, which does not seem to be absolutely correct.
This is a fiddly check to do then.

benrubson:
As indicated in #446,

benrubson:
As indicated in #446, I think we should keep :
- size_t for dataLen, as in read(2) / write(2) ;
- ssize_t as read & write return type, as in read(2) / write(2) ;
- off_t for sizes and offsets, as in stat(2).

Issue #470, opened 2018-02-20
encfs does not want to exit by itself after inactivity


benrubson:
Hello,

Sometimes, I encounter the following behavior when using inactivity auto-unmount :

```
Feb 16 21:15:07 s1 encfs/usr1[5293]: INFO Filesystem mounted: /home/usr1/dec/
Feb 16 21:23:18 s1 encfs/usr1[5293]: INFO Filesystem inactive, unmounted: /home/usr1/dec/
```

Filesystem is properly unmounted, however, `encfs` process is still running :
```
# ps aux | grep -E "^USER|encfs"
USER  PID %CPU %MEM   VSZ   RSS TT STAT STARTED    TIME COMMAND
usr1 5293  0.0  0.1 86928 55688  - Is   Fri21   1:50.86 /usr/local/bin/encfs -S -i 3 -t
    encfs/usr1 --no-default-flags -o use_ino -o direct_io /home/usr1/enc /home/usr1/dec

# lsof -p 5293
COMMAND  PID USER   FD   TYPE             DEVICE   SIZE/OFF NODE NAME
encfs   5293 usr1  cwd   VDIR      42,4090757305         24    4 /
encfs   5293 usr1  rtd   VDIR      42,4090757305         24    4 /
encfs   5293 usr1    0u  VCHR               0,23        0t0   23 /dev/null
encfs   5293 usr1    1u  VCHR               0,23        0t0   23 /dev/null
encfs   5293 usr1    2u  VCHR               0,23        0t0   23 /dev/null
encfs   5293 usr1    3u  unix 0xfffff801377976c0        0t0      ->0xfffff8017053aa20
encfs   5293 usr1    4u  VCHR               0,74 0x34bf834d   74 /dev/fuse
```

I already had this issue from time to time with 1.9.1, still the same with 1.9.4.

Any idea why ?
A race at the end of the code path ?

Many thanks 👍 

Ben

benrubson:
Going through the code path, it's as if `pthread_cond_wait()` was not returning...
https://github.com/vgough/encfs/blob/v1.9.4/encfs/main.cpp#L765

Any thoughts ?

benrubson:
I've put some debug around all `pthread_*` calls to try to understand what happens next time issue occurs.

benrubson:
So issue reproduced last night.

According to my debug log lines, idle monitoring thread is waiting here as we could expect :
[...]

PR #486, opened 2018-03-18
Changelog 1.9.5

benrubson:
Hi,

Here is a preview of the 1.9.5 changelog.
Based on the milestone :
https://github.com/vgough/encfs/milestone/4

I've moved #447 to 1.9.6, not sure if it's a good decision though.

Any thoughts ?

Many thx 👍

Ben

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -10,7 +10,7 @@ project(EncFS C CXX)
 
 set (ENCFS_MAJOR 1)
 set (ENCFS_MINOR 9)
-set (ENCFS_PATCH 4)
+set (ENCFS_PATCH 5)
 set (ENCFS_VERSION "${ENCFS_MAJOR}.${ENCFS_MINOR}.${ENCFS_PATCH}")
 set (ENCFS_SOVERSION "${ENCFS_MAJOR}.${ENCFS_MINOR}")
 set (ENCFS_NAME "Encrypted Filesystem")
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,16 @@
+v1.9.5 / 2018-03-19
+===================
+	* Correct a int/off_t buffer overflow in getSize()
+	* Correct a possible write crash
+	* Correct encfsctl cat and add reverse cat
+	* Add -c & -u cmdline options
+	* Add --noattrcache & --nodatacache cmdline options
+	* Ignore .encfs6.xml file in reverse mode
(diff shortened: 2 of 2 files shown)


benrubson:
I think we should manage to solve #470, then should be time to release 1.9.5 ?

benrubson:
Not sure if we have to wait for #470, as it may be a FreeBSD issue.
Your thoughts @rfjakob regarding the 1.9.5 release ?

benrubson:
At least the date, let me update it.
Regarding the Windows binary release, I wonder if we should wait for it to be ready ?

rfjakob:
No I would go ahead with the release now and release the windows things with 1.9.6

benrubson:
OK, PR updated.
I'll let you squash-merge it ?
Thx !

benrubson:
Please merge #516 before ?
Thx !

rfjakob:
Like this (you need a GPG key) https://github.com/rfjakob/gocryptfs/blob/12b32aa06c0475d60ee51b3753052ac2e4d09308/package-static.bash#L25

benrubson:
Why don't we simply display md5 / sha1 of the package ?

benrubson:
So to sum-up @rfjakob :
- I publish the release without attaching anything to it ;
- The previous step will then create the new tag ;
- I sync my local repository and generate the package + the .asc file ;
- I attach them to the release.

Am I right ?
Thx 👍

benrubson:
Mmmmh I think it will be better if you do the release @rfjakob, as package maintainers may already have your public key.
Here is the package description :
```
* Correct a int/off_t buffer overflow in getSize() (#468)
* Correct a possible write crash (#494)
* Correct `encfsctl cat` and add `encfsctl --reverse cat` (#483)
* Add `-c` & `-u` cmdline options (#474)
* Add `--noattrcache` & `--nodatacache` cmdline options (#500)
* Ignore .encfs6.xml file in reverse mode (#478)
* Do not count usage on root path (to avoid resetting the `--idle` option) (#471)
* Use stderr instead of stdout for error messages (#491)
* Add the ability to disable data encryption (absolutely discouraged) (#487)
* Add **Cygwin** support (#499)
```

Thx 👍

rfjakob:
Ok i will upload the tarball and the asc

rfjakob:
Released!

benrubson:
Nice job, thank you @rfjakob !

PR #487, opened 2018-03-18
 Add ability to disable file data encoding

### LINKED ISSUE #481: Encrypt filenames only / use nullCipher for data
Hi,

Would be nice to be able to encrypt filenames only.
Useful if files' content is already encrypted (whatever the method is), but goal is also to have filenames encrypted.
EncFS should then be able to use nullCipher for data itself.

Thank you 👍 

Ben

benrubson:
Hi,

This PR closes #481 adding the ability to disable file data encoding a FS creation.
Note that this must be enabled by adding the new option `--unsafe`.
And must be enforced at mount by `--unsafe`.
User then knows !

Many thx 👍 

Ben

### DIFF
--- a/encfs/FSConfig.h
+++ b/encfs/FSConfig.h
@@ -67,6 +67,8 @@ struct EncFSConfig {
   int kdfIterations;
   long desiredKDFDuration;
 
+  bool plainData;         // do not encrypt file content
+
   int blockMACBytes;      // MAC headers on blocks..
   int blockMACRandBytes;  // number of random bytes in the block header
 
@@ -79,6 +81,7 @@ struct EncFSConfig {
   EncFSConfig() : keyData(), salt() {
     cfgType = Config_None;
     subVersion = 0;
+    plainData = false;
     blockMACBytes = 0;
     blockMACRandBytes = 0;
     uniqueIV = false;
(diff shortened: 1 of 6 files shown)


benrubson:
Modifications done @rfjakob 👍 
Thank you for your review.

Regarding example filesystems, we do not have such "feature" in EncFS (yet) ?
Or perhaps I missed something ? :)
Thank you !

rfjakob:
Regarding example filesystems, you just create a filesystem and add it to the source code!

benrubson:
Perhaps this should be part of a new PR, which would provide an example per filesystem (as we don't have this yet) exhaustively ?

rfjakob:
Ok!

benrubson:
Thx 👍

rfjakob:
Travis seems unhappy, though:

> integration/reverse.t.pl .. 11/46 sh: line 1:  2269 Segmentation fault: 11  ENCFS6_CONFIG=/var/folders/bb/n7t3rs157850byt_jfdcq9k80000gn/T//encfs-reverse-tests-YOAV/plain/.encfs6.xml ./build/encfsctl cat --extpass="echo test" --reverse /var/folders/bb/n7t3rs157850byt_jfdcq9k80000gn/T//encfs-reverse-tests-YOAV/plain hello.txt > /var/folders/bb/n7t3rs157850byt_jfdcq9k80000gn/T//encfs-reverse-tests-YOAV/plain/helloenc.txt

benrubson:
Yep see #492 :-|

PR #489, opened 2018-03-18
Correct a small issue with encfsctl cat --reverse

benrubson:


benrubson:
I agree, understood for the future 👍
Thank you for your reviews @rfjakob 👍

PR #499, opened 2018-03-25
Cygwin support

benrubson:
Hi,

This PR adds support to compile and run EncFS on Cygwin.
The FUSE layer is handled by [WinFsp](https://github.com/billziss-gh/winfsp).
Port history can be found [here](https://github.com/billziss-gh/winfsp/issues/152).

All tests pass, but :
- in normal mode, symlink tests ;
- in reverse mode, absolute symlink tests.

The first point is due to how Cygwin handles symlinks. It uses small files with content beginning with a specific pattern, followed by the link target. Encryption breaks this. Perhaps we could think later on about a passthrough mode for these files.
The second point may be on WinFsp side, https://github.com/billziss-gh/winfsp/issues/153 will tell us more.
_In other words, we can read existing relative symlinks, but can't create symlinks._

See [the wiki](https://github.com/vgough/encfs/wiki) for installation instructions.

WinFsp may give nice performance improvements possibilities, [especially thanks to `READDIR_PLUS`](https://github.com/billziss-gh/winfsp/issues/44).
Some work may however be needed to use the WIN `readdir` API.

Thank you 👍 

Ben

_This PR depends on #494._

sc

### DIFF
--- a/encfs/DirNode.cpp
+++ b/encfs/DirNode.cpp
@@ -739,7 +739,9 @@ int DirNode::unlink(const char *plaintextName) {
 
   Lock _lock(mutex);
 
-  int res = 0;
+// Windows does not allow deleting opened files, so no need to check
+// There is this "issue" however : https://github.com/billziss-gh/winfsp/issues/157
+#ifndef __CYGWIN__
   if ((ctx != nullptr) && ctx->lookupNode(plaintextName)) {
     // If FUSE is running with "hard_remove" option where it doesn't
     // hide open files for us, then we can't allow an unlink of an open
@@ -747,14 +749,16 @@ int DirNode::unlink(const char *plaintextName) {
     RLOG(WARNING) << "Refusing to unlink open file: " << cyName
                   << ", hard_remove option "
                      "is probably in effect";
-    res = -EBUSY;
-  } else {
-    string fullName = rootDir + cyName;
-    res = ::unlink(fullName.c_str());
-    if (res == -1) {
-      res = -errno;
-      VLOG(1) << "unlink error: " << strerror(-res);
-    }
+    return -EBUSY;
+  }
+#endif
+
+  int res = 0;
+  string fullName = rootDir + cyName;
+  res = ::unlink(fullName.c_str());
+  if (res == -1) {
+    res = -errno;
+    VLOG(1) << "unlink error: " << strerror(-res);
   }
 
   return res;
--- a/encfs/DirNode.cpp
+++ b/encfs/DirNode.cpp
@@ -599,11 +599,26 @@ int DirNode::rename(const char *fromPlaintext, const char *toPlaintext) {
       if (renameOp) {
         renameOp->undo();
       }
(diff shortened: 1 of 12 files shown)


encfs/DirNode.cpp
```diff
@@ -739,8 +739,15 @@ int DirNode::unlink(const char *plaintextName) {
...
+  size_t maxopen = 0;
+#ifdef __CYGWIN__
+  // When deleting a file, Windows first opens it, deletes it, and closes it
+  // We then have inevitably the file opened one time 
+  maxopen = 1;
+#endif
+
   int res = 0;
-  if ((ctx != nullptr) && ctx->lookupNode(plaintextName)) {
+  if ((ctx != nullptr) && (ctx->numberOfNode(plaintextName) > maxopen)) {
```
rfjakob: Windows does not allow deleting open files, so we can just skip the check.
In other words:
```
if ( !windows && (ctx != nullptr) && (ctx->numberOfNode(plaintextName) > maxopen)) {
```

encfs/FileUtils.cpp
```diff
@@ -1739,6 +1739,9 @@ void unmountFS(const char *mountPoint) {
   // fuse_unmount does not work on Mac OS, see #428
   unmount(mountPoint, MNT_FORCE);
 #endif
+#ifdef __CYGWIN__
+  system(string("/usr/bin/pkill -f '(^|/)encfs .*/.* ").append(mountPoint).append("?( |$)'").c_str());
```
rfjakob: Shell injection vulnerabilty!

benrubson: Thank you for your review 👍 
I think a regex match of `mountPoint` would do the trick ?
Perhaps you thought about something else ?

rfjakob: The best way is to avoid the shell by using something like execl

benrubson: Done 👍

encfs/DirNode.cpp
rfjakob: I asked myself the same thing yesterday, but this code is even older than the git history, and I am not completely sure. It may be related to how encfs shares backing file descriptors between multiple open FUSE file descriptors. I believe they are stored in a table indexed by path.

encfs/FileUtils.cpp
```diff
@@ -1739,6 +1739,12 @@ void unmountFS(const char *mountPoint) {
   // fuse_unmount does not work on Mac OS, see #428
   unmount(mountPoint, MNT_FORCE);
 #endif
+#ifdef __CYGWIN__
+  if(fork() == 0)
```
benrubson: I think parent should way its child, because parent exits before `pkill` completes, and it makes some tests to fail.
I will then correct this.

benrubson:
Thank you for your review & approval @rfjakob 👍

encfs/FileUtils.cpp
benrubson: Done

PR #500, opened 2018-03-25
Add 2 options --noattrcache --nodatacache

benrubson:
Hi,

This PR adds 2 new options :
`--noattrcache : disables attr cache only.`
`--nodatacache : disables data cache only.`

So this is in addition to the existing one :
`--nocache : disables attr and data cache.`

Thank you 👍 

Ben


rfjakob:
Hmm. What is the use case here?

benrubson:
Current "problem" is that if you want to cache attributes for a small amount of time, with let's say for example `-oattr_timeout=0.1`, then data is cached without timeout.
With this PR you can then do `--nodatacache -oattr_timeout=0.1`.
I then also added `--noattrcache` for consistency.

rfjakob:
Why would you want that? Shouldn't the user just use --nocache to be safe?

rfjakob:
If there is a user need for this, we should do it, but if not, I'd suggest to keep things as simple as possible

PR #501, opened 2018-03-25
Correct a small issue in recodePath

benrubson:
Hi,

This PR corrects a small issue in recodePath, where a path to decode could begin with a trailing '+' (trailing '/' are encoded to '+').
It then revert 154839d67c9b7504687bcc2301fd2782ff9f3368 which is no more needed, this solution is much generic / much nicer.

This has been found adding `encfsctl --reverse cat`, where a user could provide a path beginning with a trailing '/'.

Thank you 👍 

Ben

### DIFF
--- a/encfs/NameIO.cpp
+++ b/encfs/NameIO.cpp
@@ -146,8 +146,8 @@ std::string NameIO::recodePath(
   string output;
 
   while (*path != 0) {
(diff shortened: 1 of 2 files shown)


benrubson:
Yep don't merge this, too many potential side effects...

PR #503, opened 2018-03-29
Retry Apple & Cygwin unmount

benrubson:
Hi,

Sometimes Cygwin process needs more than one signal to fully stop.
So this PR adds a retry loop to the `pkill` command.
While here, also add a retry loop to the Apple `unmount` command, we never know...

Thx 👍 

Ben

### DIFF
--- a/encfs/FileUtils.cpp
+++ b/encfs/FileUtils.cpp
@@ -1737,15 +1737,27 @@ void unmountFS(const char *mountPoint) {
   fuse_unmount(mountPoint, nullptr);
 #ifdef __APPLE__
   // fuse_unmount does not work on Mac OS, see #428
-  unmount(mountPoint, MNT_FORCE);
+  int ret = 3;
+  while (ret-- > 0) {
+    if (unmount(mountPoint, MNT_FORCE) == 0) {
+      break;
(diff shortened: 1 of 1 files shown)


rfjakob:
You have observed the problem on Windows? What do you see when you observe the problem?

benrubson:
Yes I have observed the problem on Windows.
Generally, mainly after some quite important activity on the FS, when unmounting (when sending `INT` signal to the process), FS correctly unmounts (`X:` - `/cygdrive/x` is not more available), but encfs process is still running.
Sending a second `INT` signal makes the process to exit.

rfjakob:
Sounds similar to https://github.com/vgough/encfs/issues/470 - I think we should rather fix the underlying issue

benrubson:
Yes sounds quite similar in terms of behavior.
Would then mean that #470 is not related to idleMonitor (I did not use it on Windows).
And that it is not FreeBSD related.

rfjakob:
Yes there may be a bug in the unmount process. Looking at the stacktraces you posted (thanks!) it's pretty strange that all threads hang in `pthread_cleanup_pop`.

PR #505, opened 2018-04-06
Exit idleMonitor thread when it has finished

benrubson:
Hi,

This PR makes the idleMonitor thread to exit as soon as it has finished.
This could help in debugging situations such as #470.

Thank you !

Ben


benrubson:
Thx for your 2 approvals @rfjakob 👍

PR #507, opened 2018-04-08
Unmount improvements

benrubson:
Hi,

I would have liked to keep some things from #503, so, this PR :
- better handles fork / child creation for Cygwin unmount ;
- uses `INT` signal [expected by WinFsp](https://github.com/billziss-gh/winfsp/wiki/SSHFS-Port-Case-Study#some-additional-challenges) rather than default `TERM`.

Thank you !

Ben

### DIFF
--- a/encfs/FileUtils.cpp
+++ b/encfs/FileUtils.cpp
@@ -1733,19 +1733,30 @@ RootPtr initFS(EncFS_Context *ctx, const std::shared_ptr<EncFS_Opts> &opts) {
 }
 
 void unmountFS(const char *mountPoint) {
-  // fuse_unmount succeeds and returns void
+  // fuse_unmount returns void, is assumed to succeed
   fuse_unmount(mountPoint, nullptr);
 #ifdef __APPLE__
   // fuse_unmount does not work on Mac OS, see #428
-  unmount(mountPoint, MNT_FORCE);
+  // However it makes encfs to hang, so we must unmount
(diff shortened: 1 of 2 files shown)


encfs/FileUtils.cpp
```diff
@@ -1740,12 +1740,15 @@ void unmountFS(const char *mountPoint) {
...
-  if(fork() == 0)
-  {
-    execl("/usr/bin/pkill", "/usr/bin/pkill", "-f", string("(^|/)encfs .*/.* ").append(mountPoint).append("?( |$)").c_str(), (char *)0);
-  }
+  pid_t pid;
   int status;
-  wait(&status);
+  if ((pid = fork()) == 0) {
+    execl("/usr/bin/pkill", "/usr/bin/pkill", "-INT", "-f", string("(^|/)encfs .*/.* ").append(mountPoint).append("?( |$)").c_str(), (char *)0);
+    _Exit(127);
```
rfjakob: Code after execl is never reached!

benrubson: Yep but this is to make the child to exit if `execl` fails 👍

benrubson:
I've just improved commit message.
Thx for your review !

encfs/FileUtils.cpp
rfjakob: Good point. But shouldn't we print an error message in that case?

benrubson: Yep, improved with log messages !

PR #509, opened 2018-04-16
Cygwin, set proper installation prefix

benrubson:
Hi,

This PR simply sets a proper installation prefix for Cygwin.
Then, binaries are next to Cygwin DLLs, which is easier for the user running EncFS out of a Cygwin console.

Thx 👍 

Ben

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -35,6 +35,10 @@ if (NOT DEFINED LIB_INSTALL_DIR)
   set (LIB_INSTALL_DIR lib)
 endif ()
 
+if (CYGWIN)
+  set (CMAKE_INSTALL_PREFIX /usr)
+endif()
+
 # We need C++ 11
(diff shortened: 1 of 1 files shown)


rfjakob:
Missing commit messagt

benrubson:
Commit message added 👍

benrubson:
Thank you @rfjakob 👍

mhogomchungu:
I think its wrong for encfs to unconditionally set ```CMAKE_INSTALL_PREFIX" variable without giving users an ability to change it. 

I wanted to change the variable to something else to easily catch all the installed files and was surprised to see my preferred installed path was ignored.

rfjakob:
@benrubson can we have set-if-not-set-by-user ?

benrubson:
#513 should help !

benrubson:
> I think its wrong for encfs to unconditionally set CMAKE_INSTALL_PREFIX variable without giving users an ability to change it.

Should be OK now @mhogomchungu 👍

PR #510, opened 2018-04-16
Cygwin, support Windows-style parameter paths

benrubson:
Hi,

This PR enables the use of Windows-style paths as parameters.
Now, user can use for example :
`encfs C:\cipher X:`

Thank you 👍 

Ben


benrubson:
So as you can see here I had to create a new internal configuration item, `unmountPoint`, to retain mount point as originally provided by the user.
In Cygwin, `pkill` is used to unmount, and in order to identify the correct `encfs` process to stop (one could mount several FS at the same time), `pkill` looks at the whole `encfs` command line for the given mount point.
Unfortunately, the existing internal configuration item `mountPoint` is updated (from Win to Linux if needed, and with a trailing slash), so can't be used to unmount, as it would have chances not to match existing processes.

I did not found an easier/shorter way to solve this issue.

Issue #508, opened 2018-04-12
A request for a binary version of encfs for windows to be uploaded


mhogomchungu:
This[1] pull request made it possible to run encfs on windows.

I have a project called SiriKali[2] and would like to add support for encfs on windows and will appreciate very much for a binary version of encfs for windows to be uploaded to give some of us an opportunity to play with it.

I plan to make a new release of SiriKali on the first of April, 2018 and i'll need something to play with much sooner than that date.

[1] https://github.com/vgough/encfs/pull/499
[2] http://mhogomchungu.github.io/sirikali

benrubson:
@mhogomchungu, to be as fast as possible, as a first solution, did you try to compile it by yourself ?
It's rather easy, as per [the Wiki instructions](https://github.com/vgough/encfs/wiki/Cygwin).

For runtime, you should be able to run the compiled binary out of the Cygwin files' tree.
You will however require some DLLs such as `cygwin1.dll` : missing ones will be notified when you'll start `encfs`. Just copy them from Cygwin file's tree next to the binary, it should do the trick 👍
And of course you'll still need WinFsp, as the "FUSE layer".

Not sure how a Windows binary would have to be distributed ?
- directly as a Cygwin package, among all the other Cygwin packages ?
- here in EncFS repository, with required DLLs ? As both 32 and 64bits ?
- something else ?

Every solution would require for sure some additional maintenance.

mhogomchungu:
Best thing in my opinion is to distribute the binary here as one of the released files.

Best solution will be to statically build everything to end up with a single executable. This way,the user can put the executable anywhere on their system is just use it.

Second best thing is to ship the binary and all of its dependencies in a folder. An installer will be best used here. 

With both alternatives, it will probably be necessary to ship 32 bit and 64 bit packages.

Will see if i can build it locally.

mhogomchungu:
I managed to build it and i think i found a bug

```
C:\encfs-master\build\encfs>encfs -f C:\cipher Z:
A drive is prefered for mouting, so a path like /cygdrive/x should rather be used. Mounting anyway.
EncFS Password:
Cannot create WinFsp-FUSE file system: invalid mount point.

C:\encfs-master\build\encfs>
```
A drive letter of something like ```Z:``` is rejected as it wants something like```Z```. I think the first way is the correct way when writing drive letters.

benrubson:
Simply use `/cygdrive/z` as mount point, as proposed 👍

[...]

PR #515, opened 2018-04-26
Add a windows installer script

mhogomchungu:
This pull requests adds an "inno setup"[1] installer script to the project. This script will make it possible to easily package and distribute encfs for windows using an installer.

[1] http://www.jrsoftware.org/isinfo.php


windows-installer.iss
```diff
@@ -0,0 +1,92 @@
...
+; SEE THE DOCUMENTATION FOR DETAILS ON CREATING INNO SETUP SCRIPT FILES!
+
+#define MyAppName "Encfs"
+#define MyAppVersion "1.9.5"
+#define MyAppPublisher "encfs"
+#define MyAppURL "https://github.com/vgough/encfs"
+#define MyAppExeName "encfs.exe"
+
+; This variable should contain a path to where libraries needed by encfs are installed.
+#define MyLibraryPath "C:\cygwin\usr\bin"
```
benrubson: Perhaps drop the `usr` part of the path ?
As the real Windows path is `C:\cygwin\bin`, `C:\cygwin\usr\bin` being a mount point done by Cygwin to `C:\cygwin\bin`.

windows-installer.iss
```diff
@@ -0,0 +1,92 @@
...
+
+; This variable should contain a path to where cygwin binaries to be packaged are located.
+#define MyEXEPath "C:\cygwin\usr\bin"
```
benrubson: Same here, perhaps drop the `usr` part of the path which may be not available out of Cygwin itself ?
Thx 👍

windows-installer.iss
mhogomchungu: Done on both recommendations.

benrubson: Thank you 👍

benrubson:
Is assume this script is rather dedicated to the 32 bits version ?
(as Cygwin 64 bits is by default installed into `C:\cygwin64`)

We could also simply assume to install the 64 bits version into `C:\cygwin`...

Sounds like Appveyor has both 32 and 64 bits Cygwin versions built-in :
https://www.appveyor.com/docs/build-environment/#mingw-msys-cygwin
We could simply rename `C:\cygwin64` to `C:\cygwin` in an automated build scenario.

mhogomchungu:
I think a decision will have to be made here.

1. Ship only 32 bit version of the installer because it works with both 32 bit and 64 bit windows.

2. Ship installers for both 32 bit and 64 bit installers and let users pick the one they prefer.

If we go with the former, nothing needs to be done, if with the latter then we could do the folder substitution trick as suggested or have a separate installer script for 64 bit and name it differently to differentiate the two. Having two installer script will make it easier to build both installers with the same CI instance if it will be possible to have versions of cygwin installed.

benrubson:
Because of the issue we discussed [here](https://github.com/vgough/encfs/issues/508#issuecomment-384597646), I think we need at least a 64 bits version, so that the 32 bits version should be chosen by users running a 32 bits OS only.

> Having two installer script will make it easier to build both installers with the same CI instance if it will be possible to have versions of cygwin installed.

Yes you're right. A little bit (just a little) more difficult to maintain though, if we need to update them (did you update both scripts ?).
According to what I understand [from this](https://www.appveyor.com/docs/build-environment/#mingw-msys-cygwin), both 32 and 64 bits Cygwin versions are installed in all AppVeyor images.

billziss-gh:
FWIW, my approach has been to package all necessary Cygwin files within the installer in [SSHFS-Win](https://github.com/billziss-gh/sshfs-win) and [NFS-Win](https://github.com/billziss-gh/nfs-win).

benrubson:
We should too, and according to https://github.com/billziss-gh/winfsp/issues/161, sounds like it could be important to test and choose Cygwin version to package !

mhogomchungu:
So, whats the verdict on the pull request?

This is as far as i can take take the installer script. Decisions on how many packages to build and from where could be handled internally later on.

rfjakob:
Thanks for the installer script! I have merged it into the new "cygwin" branch as 54f8aa1 . We should experiment there, get the build infrastructure right, and squash and merge into master when ready.

PR #517, opened 2018-04-27
Add signing hint in package-source.sh helper

benrubson:
Hi,

This PR adds the output of archive hashes in package-source.sh helper.

Thx 👍 

Ben


rfjakob:
Uhmm why?

rfjakob:
We have never published hashes, see https://github.com/vgough/encfs/releases

benrubson:
I've then updated this to just add the signing hit.

PR #521, opened 2018-05-12
Handle CBC padding

### LINKED ISSUE #9: Stream Cipher Used to Encrypt Last File Block
From: https://defuse.ca/audits/encfs.htm

Exploitability: **Unknown**
Security Impact: **High**

  As reported in [[1](http://archives.neohapsis.com/archives/fulldisclosure/2010-08/0316.html)], EncFS uses a stream cipher mode to encrypt the last file block. The change log says that the ability to add random bytes to a block was added as a workaround for this issue. However, it does not solve the problem, and is not enabled by default.

  EncFS needs to use a block mode to encrypt the last block.

  EncFS's stream encryption is unorthodox:

```
1. Run "Shuffle Bytes" on the plaintext.
    N[J+1] = Xor-Sum(i = 0 TO J) { P[i] }
    (N = "shuffled" plaintext value, P = plaintext)
2. Encrypt with (setIVec(IV), key) using CFB mode.
3. Run "Flip Bytes" on the ciphertext.
    This reverses bytes in 64-byte chunks.
4. Run "Shuffle Bytes" on the ciphertext.
5. Encrypt with (setIVec(IV + 1), key) using CFB mode.

Where setIVec(IV) = HMAC(globalIV || (IV), key), and,
    - 'globalIV' is an IV shared across the entire filesystem.
    - 'key' is the encryption key.
```

  This should be removed and replaced with something more standard. As far as I can see, this provides no useful security benefit, however, it is relied upon to prevent the attacks in [[1](http://archives.neohapsis.com/archives/fulldisclosure/2010-08/0316.html)]. This is security by obscurity.

Edit : [1] may be unavailable, so here it is from archives.org :
```
[Full-disclosure] Multiple Vulnerabilities in EncFS
From: Micha Riser (micha[at]povworld.org)
Date: Thu Aug 26 2010 - 07:05:18 CDT
(...)
3. Last block with single byte is insecure 
------------------------------------------------------- 
The CFB cipher mode is insecure if it is used twice with the same 
initialization vector. In CFB, the ﬁrst block of the plain text is XOR-ed with 
the encrypted IV: 
  C0 = P0 XOR Ek (IV ) 
Therefore, for two cipher blocks C0 and C0' encrypted with the same IV, it 
holds that: 
  C0 XOR C0' = (P0 XOR Ek (IV )) XOR (P0' XOR Ek (IV )) = P0 XOR P0' 
This means that an attacker gets the XOR of the two plain texts. EncFs uses a 
modified version of CFB which additionally shuffles and reverses bytes. It is not 
clear however, if the modiﬁcations generally help against this problem. 

A security problem arises deﬁnitely if the last block contains only a single 
byte and an attacker has two versions of the last block. Operating on a single 
byte, the shuﬄe and reverse operation do nothing. What remains is a double 
encryption with CFB and XOR-ing the two cipher bytes gives the XOR of the 
two plain text bytes due to the reason described above. Encrypting the last 
block with a stream cipher instead of a block cipher saves at most 16 bytes 
(one cipher block). We think it would be better to sacriﬁce these bytes and in 
exchange rely only on a single encryption mode for all blocks which simpliﬁes 
both the crypto analysis and the implementation.
```

### LINKED ISSUE #10: Generating Block IV by XORing Block Number
From: https://defuse.ca/audits/encfs.htm  

Exploitability: **Low**
Security Impact: **Medium**

Given the File IV (an IV unique to a file), EncFS generates per-block IVs by XORing the File IV with the Block Number, then passing the result to setIVec(), which is described in #9. This is not a good solution, as it leads to IV re-use when combined with the last-block stream cipher issue in #9:

 The stream algorithm (see previous section) adds 1 to the IV, which could _undo_ the XOR with the block number, causing the IV to be re-used. Suppose the file consists of one and a half blocks, and that the File IV's least significant bit (LSB) is 1. The first block will be encrypted with the File IV (block number = 0). The second (partial) block will be encrypted with File IV XOR 1 (since block number = 1), making the LSB 0, using the stream algorithm.  The stream algorithm adds 1 to the IV, bringing the LSB back to 1, and hence the same IV is used twice. The IVs are reused with different encryption modes (CBC and CFB), but CFB mode starts out similar to CBC mode, so this is worrisome.

EncFS should use a mode like XTS for random-access block encryption.

**Correction 12/05/2014**:
XTS mode is probably not the ideal option, see Thomas Ptacek's [blog post](http://sockpuppet.org/blog/2014/04/30/you-dont-want-xts/) for good reasons why.

benrubson:
Hi,

This PR handles CBC padding, to get rid of cipher stream mode in the last data block.
This then closes #9 (and as a result closes #10), thus increasing EncFS security level.

Padding scheme is the one recently discussed in #9, and is detailed in the first commit of this PR.

Of course CBC padding is enabled for newly created filesystems, older ones will still use stream mode, for compatibility purpose.

I'm working on reverse write, but all other modes (read/write in normal mode, read in reverse mode) are already working, so that you can test & review the achieved work.

Thank you :+1: 

Ben

### DIFF
--- a/encfs/BlockFileIO.cpp
+++ b/encfs/BlockFileIO.cpp
@@ -43,7 +43,7 @@ static void clearCache(IORequest &req, unsigned int blockSize) {
 BlockFileIO::BlockFileIO(unsigned int blockSize, const FSConfigPtr &cfg)
     : _blockSize(blockSize), _allowHoles(cfg->config->allowHoles) {
   CHECK(_blockSize > 1);
-  _cache.data = new unsigned char[_blockSize + 1]; // we will need some room for padding
+  _cache.data = new unsigned char[_blockSize + 16]; // we will need some room for padding
   _noCache = cfg->opts->noCache;
 }
 
--- a/encfs/BlockFileIO.cpp
+++ b/encfs/BlockFileIO.cpp
@@ -45,6 +45,13 @@ BlockFileIO::BlockFileIO(unsigned int blockSize, const FSConfigPtr &cfg)
   CHECK(_blockSize > 1);
   _cache.data = new unsigned char[_blockSize + 16]; // we will need some room for padding
   _noCache = cfg->opts->noCache;
+  /* Even if cache is disabled, we may need to cache some data.
+   * This is the case in reverse mode, with CBC padding, partially writing the last block.
+   * If a write request is made with only some of the first bytes of the last block,
+   * the last bytes of this request won't certainly be properly written to disk (because of CBC mode).
+   * We must then cache these bytes so that subsequent request, which will certainly complete
+   * this last block, will properly retrieve them and finally write the full (or correctly padded) block. */
+  _cachePartialWrite = cfg->reverseEncryption && cfg->config->padding;
 }
 
 BlockFileIO::~BlockFileIO() {
@@ -65,9 +72,9 @@ ssize_t BlockFileIO::cacheReadOneBlock(const IORequest &req) const {
   /* we can satisfy the request even if _cache.dataLen is too short, because
    * we always request a full block during reads. This just means we are
    * in the last block of a file, which may be smaller than the blocksize.
-   * For reverse encryption, the cache must not be used at all, because
+   * For reverse encryption, the cache should not be used at all, because
    * the lower file may have changed behind our back. */
-  if ((!_noCache) && (req.offset == _cache.offset) && (_cache.dataLen != 0)) {
+  if ((_cache.dataLen != 0) && (req.offset == _cache.offset)) {
     // satisfy request from cache
     size_t len = req.dataLen;
     if (_cache.dataLen < len) {
@@ -76,23 +83,25 @@ ssize_t BlockFileIO::cacheReadOneBlock(const IORequest &req) const {
     memcpy(req.data, _cache.data, len);
     return len;
   }
-  if (_cache.dataLen > 0) {
-    clearCache(_cache, _blockSize);
-  }
 
-  // cache results of read -- issue reads for full blocks
-  IORequest tmp;
-  tmp.offset = req.offset;
-  tmp.data = _cache.data;
-  tmp.dataLen = _blockSize;
-  ssize_t result = readOneBlock(tmp);
+  // issue reads for full blocks, and keep result in cache if we're allowed to
+  _cache.offset = req.offset;
+  _cache.dataLen = _blockSize;
+  ssize_t result = readOneBlock(_cache);
   if (result > 0) {
-    _cache.offset = req.offset;
     _cache.dataLen = result;  // the amount we really have
     if ((size_t)result > req.dataLen) {
       result = req.dataLen;  // only as much as requested
     }
     memcpy(req.data, _cache.data, result);
+    if (!_noCache) {
+      // zero the last bytes of the cache
+      memset(_cache.data + _cache.dataLen, 0, _blockSize - _cache.dataLen);
+    } else {
+      clearCache(_cache, _blockSize);
+    }
+  } else {
+    clearCache(_cache, _blockSize);
   }
   return result;
 }
@@ -101,19 +110,16 @@ ssize_t BlockFileIO::cacheWriteOneBlock(const IORequest &req) {
   // Let's point request buffer to our own buffer, as it may be modified by
   // encryption : originating process may not like to have its buffer modified
   memcpy(_cache.data, req.data, req.dataLen);
-  IORequest tmp;
-  tmp.offset = req.offset;
-  tmp.data = _cache.data;
-  tmp.dataLen = req.dataLen;
-  ssize_t res = writeOneBlock(tmp);
-  if (res < 0) {
-    clearCache(_cache, _blockSize);
-  }
-  else {
-    // And now we can cache the write buffer from the request
+  _cache.offset = req.offset;
+  _cache.dataLen = req.dataLen;
+  ssize_t res = writeOneBlock(_cache);
+  // we cache if we're allowed to or if we are writing the last partial block in CBC reverse mode
+  if ((res > 0) && (!_noCache || (_cachePartialWrite && (req.dataLen < _blockSize)))) {
     memcpy(_cache.data, req.data, req.dataLen);
-    _cache.offset = req.offset;
-    _cache.dataLen = req.dataLen;
+    // zero the last bytes of the cache
+    memset(_cache.data + req.dataLen, 0, _blockSize - req.dataLen);
+  } else {
+    clearCache(_cache, _blockSize);
   }
   return res;
 }
--- a/encfs/BlockFileIO.cpp
+++ b/encfs/BlockFileIO.cpp
@@ -43,7 +43,7 @@ static void clearCache(IORequest &req, unsigned int blockSize) {
 BlockFileIO::BlockFileIO(unsigned int blockSize, const FSConfigPtr &cfg)
     : _blockSize(blockSize), _allowHoles(cfg->config->allowHoles) {
   CHECK(_blockSize > 1);
-  _cache.data = new unsigned char[_blockSize];
+  _cache.data = new unsigned char[_blockSize + 1]; // we will need some room for padding
   _noCache = cfg->opts->noCache;
 }
 
--- a/encfs/CipherFileIO.cpp
+++ b/encfs/CipherFileIO.cpp
@@ -635,10 +635,10 @@ int CipherFileIO::truncate(off_t size) {
     }
     reopen = 1;
   }
-  if (!haveHeader) {
+  if (!haveHeader && !haveCBCPadding) {
     res = BlockFileIO::truncateBase(size, base.get());
   } else {
-    if (0 == fileIV) {
+    if (haveHeader && (0 == fileIV)) {
       // empty file.. create the header..
       res = initHeader();
     }
@@ -648,7 +648,13 @@ int CipherFileIO::truncate(off_t size) {
       res = BlockFileIO::truncateBase(size, nullptr);
     }
     if (res == 0) {
-      res = base->truncate(size + HEADER_SIZE);
+      if (haveCBCPadding && (size > 0)) {
+        size += (size - 1) / blockSize() + fsConfig->cipher->cipherBlockSize();
+      }
+      if (haveHeader) {
+        size += HEADER_SIZE;
+      }
+      res = base->truncate(size);
     }
   }
   if (reopen == 1) {
--- a/encfs/CipherFileIO.cpp
+++ b/encfs/CipherFileIO.cpp
@@ -467,7 +467,7 @@ ssize_t CipherFileIO::readOneBlock(const IORequest &req) const {
                      blockNum ^ fileIV);  // cast works because we work on a
                                           // block and blocksize fit an int
       if (ok) {
-        if (readSize < req.dataLen) {
+        if ((size_t)readSize < req.dataLen) {
           // this is the last block, we must pad it with cipherBlockSize() bytes
           memset(tmpReq.data + readSize, 0x00, cbs - padBytes);
           readSize += cbs - padBytes;
@@ -531,6 +531,8 @@ ssize_t CipherFileIO::writeOneBlock(const IORequest &req) {
   unsigned int bs = blockSize();
   off_t blockNum = req.offset / bs;
 
+  ssize_t writeSize = req.dataLen;
+
   if (haveHeader && (fileIV == 0)) {
     int res = initHeader();
     if (res < 0) {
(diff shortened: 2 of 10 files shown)


rfjakob:
Stupid question: How are your Go skills? I am thinking about whether we should use the gocryptfs test suite for EncFS as well.

benrubson:
I've never worked with Go yet, so I would have to learn.
Should not be too difficult, but I'm not sure I'll have enough time for this during the next weeks / months.
What sort of tests does the gocryptfs test suite run ?
Would it be easy, without too much work, to run it over EncFS ?
What about rewriting it into Perl for example ?

rfjakob:
Most interesting is probably
* https://github.com/rfjakob/gocryptfs/blob/master/tests/matrix/matrix_test.go
* https://github.com/rfjakob/gocryptfs/blob/master/tests/reverse/correctness_test.go

Rewrite in Perl would be possible, but I'm not sure if it's worth it. Perl does not have much of a future IMO.

Having the test suite in C++ seems pretty painful.

Having the test suite in Go instead of Perl? Would enable sharing with gocryptfs. For me it would be a lot easier to work with obviously. Question is whether Go is easier to work with even if you have some Perl experience?

rfjakob:
Maybe adding EncFS support to https://github.com/rfjakob/fuse-xfstests would be enough?

benrubson:
I agree regarding C++.
Go should not be too hard to work with, however I'm thinking about compatibility with some systems such as Cygwin, but I just saw there's a Windows Go version.
There are certainly some EncFS tests that are not in the gocryptfs suite, and vice versa.
Are there many gocryptfs tests not in EncFS ? Or other things which make you think sharing gocryptfs suite would be nice ?

> Having the test suite in Go instead of Perl? Would enable sharing with gocryptfs. For me it would be a lot easier to work with obviously.

A lot easier for you to work with Go rather than with Perl ?
You would not then consider a shared suite rewritten in Perl ?

> Maybe adding EncFS support to https://github.com/rfjakob/fuse-xfstests would be enough?

You also use xfstests in addition to the gocryptfs test suite ?

rfjakob:
> A lot easier for you to work with Go rather than with Perl ? You would not then consider a shared suite rewritten in Perl ?

Yes, Go is a lot easier. No more Perl! ;)

> You also use xfstests in addition to the gocryptfs test suite ?

Yes

rfjakob:
> Are there many gocryptfs tests not in EncFS ? Or other things which make you think sharing gocryptfs suite would be nice ?

The gocryptfs test suite ich much bigger, yes. But many cases are probably also covered by xfstests, maybe indirectly.

benrubson:
> The gocryptfs test suite is much bigger, yes. But many cases are probably also covered by xfstests, maybe indirectly.

Perhaps we could try at least to have `xfstests` running over EncFS.

And also make an exhaustive list of tests that gocryptfs suite runs, so that we can add them to the (Perl) tests done by EncFS, and vice versa.
And when we modify tests in one of the 2 projects, we could think about proposing the modification to the other project, so that they keep synced.
What do you think about this ?
Tests are not modified every day, so should not be so painful.

encfs/CipherFileIO.cpp
```diff
@@ -168,18 +168,33 @@ int CipherFileIO::getAttr(struct stat *stbuf) const {
   // stat() the backing file
```
rfjakob: Commit message is missing for commit `Padding, size` !

encfs/CipherFileIO.cpp
```diff
@@ -168,18 +168,33 @@ int CipherFileIO::getAttr(struct stat *stbuf) const {
...
   int res = base->getAttr(stbuf);
```
rfjakob: The change adds a few cases to the "if maze". Please simplify the whole thing a bit by doing

    if(res != 0)
        return res;
    if(!S_ISREG(stbuf->st_mode))
       // special files don't need size adjustment.
       return 0;
    if(stbuf->st_size == 0)
        // empty files stay empty
       return 0;

here. In other words, handle the simple cases first and early return.

encfs/CipherFileIO.cpp
```diff
@@ -168,18 +168,33 @@ int CipherFileIO::getAttr(struct stat *stbuf) const {
...
+      	/* Instead of using rAssert, we could also relax the rule and return 0.
+      	 * Think about a file which would have been partially written.
+      	 * Same for getSize() below. */
+        rAssert(stbuf->st_size >= HEADER_SIZE);
+        stbuf->st_size -= HEADER_SIZE;
+      }
+      if (haveCBCPadding && (stbuf->st_size > 0)) {
+        rAssert(stbuf->st_size >= fsConfig->cipher->cipherBlockSize());
+        stbuf->st_size -= fsConfig->cipher->cipherBlockSize();
+        stbuf->st_size -= stbuf->st_size / (blockSize() + 1);
```
rfjakob: This is rounding of some kind? Needs a comment, or better, it's own function. In gocryptfs I have `cipherSizeToPlainSize()` and `plainSizeToCipherSize()`.

encfs/CipherFileIO.cpp
```diff
@@ -194,11 +209,22 @@ off_t CipherFileIO::getSize() const {
...
+  if (!fsConfig->reverseEncryption) {
+    if (haveHeader && (size > 0)) {
       rAssert(size >= HEADER_SIZE);
       size -= HEADER_SIZE;
-    } else {
+    }
+    if (haveCBCPadding && (size > 0)) {
+      rAssert(size >= fsConfig->cipher->cipherBlockSize());
+      size -= fsConfig->cipher->cipherBlockSize();
+      size -= size / (blockSize() + 1);
```
rfjakob: This duplicates the logic from getAttr(), right? Why?

encfs/BlockFileIO.cpp
```diff
@@ -43,7 +43,7 @@ static void clearCache(IORequest &req, unsigned int blockSize) {
 BlockFileIO::BlockFileIO(unsigned int blockSize, const FSConfigPtr &cfg)
```
rfjakob: Commit message is missing

encfs/CipherFileIO.cpp
```diff
@@ -383,26 +383,73 @@ int CipherFileIO::generateReverseHeader(unsigned char *headerBuf) {
...
   off_t blockNum = req.offset / bs;
 
   IORequest tmpReq = req;
 
+  // adjust offset and length if we pad
+  if (haveCBCPadding && !fsConfig->reverseEncryption) {
+    tmpReq.offset += tmpReq.offset / tmpReq.dataLen;
+    tmpReq.dataLen += 1;
+  }
+
```
rfjakob: You use the remainder of the division of the offset by the request length.... Why? What does this do?

encfs/CipherFileIO.cpp
```diff
@@ -483,36 +483,92 @@ ssize_t CipherFileIO::writeOneBlock(const IORequest &req) {
...
+    if (!fsConfig->reverseEncryption) {
+      // we need padding bytes so that dataLen is a multiple of cipherBlockSize()
+      padBytes = cbs - (req.dataLen % cbs);
+      // we then add the first padding byte : 0x80
+      req.data[req.dataLen] = 0x80;
+      // if needed, we add the other padding bytes : 0x00
+      memset(req.data + req.dataLen + 1, 0x00, padBytes - 1);
+      ok = blockWrite(req.data, (int)req.dataLen + padBytes,
+                      blockNum ^ fileIV);  // cast works because we work on a
+                                           // block and blocksize fit an int
```
rfjakob: On error, please return early to cut down "if maze". In other words,

    if(!ok) {
    log error
    return
    }

encfs/CipherFileIO.cpp
```diff
@@ -635,10 +635,10 @@ int CipherFileIO::truncate(off_t size) {
     }
```
rfjakob: Commit message is missing

encfs/CipherFileIO.cpp
```diff
@@ -648,7 +648,13 @@ int CipherFileIO::truncate(off_t size) {
       res = BlockFileIO::truncateBase(size, nullptr);
     }
     if (res == 0) {
-      res = base->truncate(size + HEADER_SIZE);
+      if (haveCBCPadding && (size > 0)) {
+        size += (size - 1) / blockSize() + fsConfig->cipher->cipherBlockSize();
```
rfjakob: Looks like another user for `cipherSizeToPlainSize()`

encfs/CipherFileIO.cpp
```diff
@@ -391,14 +391,23 @@ ssize_t CipherFileIO::readOneBlock(const IORequest &req) const {
   IORequest tmpReq = req;
```
rfjakob: Commit message is missing

integration/reverse.t.pl
```diff
@@ -204,7 +204,7 @@ sub largeRead {
     my $cname = encName("largeRead");
     # cfh ... ciphertext file handle
     ok(open(my $cfh, "<", "$ciphertext/$cname"), "open ciphertext largeRead file");
-    ok(sizeVerify($cfh, 1024*1024), "1M file size");
+    ok(sizeVerify($cfh, (1024 * 1024) + int((1024 * 1024 - 1) / 1023) + 16), "1M file size");
```
rfjakob: We need to test both the old AND the new on-disk format!

rfjakob:
PS: I reviewed each commit separately, not all changes at once. Please view the comments using the [commit view](https://github.com/vgough/encfs/pull/521/commits/1714d43ba5182b91b194c0862a0e6b6228fc45b8) and click "next" on the top right to see the next one

encfs/CipherFileIO.cpp
benrubson: Here, `tmpReq.dataLen == blockSize` (each read / write request is `blockSize` bytes in length).
Remember that every block is padded with 1 byte.
We then add to the requested offset the number of backing padding bytes which exist before the upper position we want to read from.

benrubson:
Thank you for your review @rfjakob, and sorry for my delayed answered, I was stuck on other things...
I'm back on this, I pushed 4 commits related to your comments. Feel free if needed 👍 
I now resume the work on reverse write to complete this PR.

benrubson:
Holes' support to complete, and the PR should be OK 👍

benrubson:
Still some issues with truncate in reverse mode, and holes to fully support.

encfs/CipherFileIO.cpp
```diff
@@ -48,9 +48,35 @@ static Interface CipherFileIO_iface("FileIO/Cipher", 2, 0, 1);
...
+    }
+    return Padding_Normal;
+  }
+  return Padding_Disabled;
+}
+
 CipherFileIO::CipherFileIO(std::shared_ptr<FileIO> _base,
                            const FSConfigPtr &cfg)
-    : BlockFileIO(cfg->config->blockSize, cfg),
+    : BlockFileIO(cfg->config->blockSize - (checkCBCPadding(cfg) == Padding_Normal ? 1 : 0), cfg),
```
rfjakob: Needs a comment (why do we do this?)

encfs/CipherFileIO.cpp
```diff
@@ -127,6 +154,49 @@ bool CipherFileIO::setIV(uint64_t iv) {
   return base->setIV(iv);
 }
 
+/**
+ * Helper function to adjust plain size to cipher size
+ * according to configuration (file header, padding).
+ */
+void CipherFileIO::plainSizeToCipherSize(off_t *size) const {
```
rfjakob: It's good that it is now a separate function. This means it can get a unit test. I think it needs one!

encfs/CipherFileIO.cpp
```diff
@@ -127,6 +154,49 @@ bool CipherFileIO::setIV(uint64_t iv) {
...
+      *size += HEADER_SIZE;
+    }
+  }
+}
+
+/**
+ * Helper function to adjust cipher size to plain size
+ * according to configuration (file header, padding).
+ */
+void CipherFileIO::cipherSizeToPlainSize(off_t *size) const {
```
rfjakob: Needs unit test.

encfs/CipherFileIO.cpp
```diff
@@ -326,26 +389,119 @@ int CipherFileIO::generateReverseHeader(unsigned char *headerBuf) {
...
+  int cbs = fsConfig->cipher->cipherBlockSize();
   int bs = blockSize();
   off_t blockNum = req.offset / bs;
 
   IORequest tmpReq = req;
 
+  // adjust offset and length if we pad
+  if (haveCBCPadding) {
+    if (!fsConfig->reverseEncryption) {
+      tmpReq.offset += tmpReq.offset / tmpReq.dataLen;
```
rfjakob: What does this line do? Needs a comment.

encfs/CipherFileIO.cpp
```diff
@@ -326,26 +389,119 @@ int CipherFileIO::generateReverseHeader(unsigned char *headerBuf) {
...
+    if (!fsConfig->reverseEncryption) {
+      // remove the padding bytes which could have been added after the ciphertext
+      int padOutBytes = readSize % cbs;
+      readSize -= padOutBytes;
+      if (readSize > 0) {
+        ok = blockRead(tmpReq.data, (int)readSize,
+                       blockNum ^ fileIV);  // cast works because we work on a
+                                            // block and blocksize fit an int
+        if (ok) {
+          // we could have some padding bytes at the end of the plain data (X * 0x00)
```
rfjakob: Do you mean "ciphertext data" ?

encfs/CipherFileIO.cpp
```diff
@@ -326,26 +389,119 @@ int CipherFileIO::generateReverseHeader(unsigned char *headerBuf) {
...
+  if (haveHeader && (fileIV == 0) && (readSize > 0)) {
+    int res = const_cast<CipherFileIO *>(this)->initHeader();
+    if (res < 0) {
+      return res;
+    }
+  }
+
+  bool ok = true;
+
+  if (haveCBCPadding && (readSize >= 0)) {
```
rfjakob: Please return early when readSize <= 0:

```
if (readSize <= 0)
return 0;
```

encfs/CipherFileIO.cpp
```diff
@@ -326,26 +389,119 @@ int CipherFileIO::generateReverseHeader(unsigned char *headerBuf) {
...
+      }
+      if ((readSize > 0) && ((bs - readSize) < cbs)) {
+        readSize = cbs - (bs - readSize);
+        memset(tmpReq.data, 0x00, readSize);
+      } else {
+        readSize = 0;
+        VLOG(1) << "readSize zero for offset " << req.offset;
       }
     }
+  }
```
rfjakob: Hmm. I feel like this has way too many levels of indentation. Maybe split the subsections into functions? Or at least add comments for what block starts at each "else {" ?

encfs/CipherFileIO.cpp
```diff
@@ -377,42 +535,127 @@ ssize_t CipherFileIO::writeOneBlock(const IORequest &req) {
...
+
+  if (!ok) {
     VLOG(1) << "encodeBlock failed for block " << blockNum << ", size "
             << req.dataLen;
-    res = -EBADMSG;
+    writeSize = -EBADMSG;
   }
-  return res;
+
+  if (writeSize > 0) {
```
rfjakob: please check "writeSize <= 0" early and return out. Gets rid of one level of indentation.

encfs/FileUtils.cpp
```diff
@@ -1037,10 +1050,11 @@ RootPtr createV6Config(EncFS_Context *ctx,
   int blockMACBytes = 0;        // selectBlockMAC()
   int blockMACRandBytes = 0;    // selectBlockMAC()
   bool plainData = false;       // selectPlainData()
+  bool padding = true;          // selectPadding()
   bool uniqueIV = true;         // selectUniqueIV()
   bool chainedIV = true;        // selectChainedIV()
   bool externalIV = false;      // selectExternalChainedIV()
-  bool allowHoles = true;       // selectZeroBlockPassThrough()
+  bool allowHoles = false;       // selectZeroBlockPassThrough()
```
rfjakob: Woah, that's big change. Why no file holes anymore?

benrubson:
_Unfortunately I'm closing this, feel free to continue / finish it, was not so far from the end AFAIR._

PR #583, opened 2020-02-24
Clean & harden normal tests

benrubson:
Hi,

I widely cleaned normal tests, and also hardened them.

For example, I now check / wait that FS is mounted after mount.
And dismounted after unmount.
This makes the normal tests to correctly run on FreeBSD CI :
https://cirrus-ci.com/task/4619255987568640

This should also help other OS.

Thx 👍


rfjakob:
Did you run some kind of code formatter here? There's almost no line unchanged!

benrubson:
You're rather right, I mainly re-organized `normal.t.pl` file, making it easier to read.

I then added a test after mount, to wait for FS to be mounted.
And a test after unmount, to wait for the FS to be unmounted.
This is the main modification, which makes normal tests to correctly run on FreeBSD CI.

The other modification is the file creation test, I modified :
`qx(date > "$crypt/df.txt");`
into :
`system("cat $0 > $decrypted/create");`
The test is then :
`system("diff $0 $decrypted/create");`

So, in addition to be "FreeBSD compliant", this file is easier to read, it will be easier to modify in the future to add new tests etc...

PR #584, opened 2020-02-25
Remove confusing LARGEFILE warning

benrubson:
Hi,

This PR removes a confusing & not needed compilation warning.

Thx 👍


rfjakob:
Hmm, isn't it bad when O_LARGEFILE is not supported? Should we throw an error instead of a warning?

benrubson:
No, as we use compilation option `_FILE_OFFSET_BITS=64`.
In addition, where `O_LARGEFILE` is used in the code, it is simply passed-through, not explicitly used.
So there's no reason to get `O_LARGEFILE` if it's not supported on the system...

PR #595, opened 2020-03-01
Add Windows CI

benrubson:
Hi,

Good news, I managed to get a working Windows CI.
Some integration tests are failing, as expected, but the whole process is OK 👍

Ben

PR #597, opened 2020-03-03
Reorganize CI files

benrubson:
Hi,

As discussed in #595, here's a PR which reorganize CI files, for easier upcoming updates.

Thank you 👍 

Ben


benrubson:
_#583 should be merged first so that FreeBSD tests won't fail / timeout._

benrubson:
This now builds / tests correctly, thus you should be able to approve it @rfjakob.
Many thx 👍

.gitattributes
```diff
@@ -0,0 +1 @@
+* text=auto eol=lf
```
rfjakob: I think this should be in the Windows PR?

benrubson: I've then removed it, I'll add it into Windows PR.
Thank you for your review 👍

benrubson:
Let's approve this if you have some seconds @rfjakob ?
Many thanks 👍

PR #673, opened 2024-12-04
encfs: Update LibreSSL compat

orbea:
As of LibreSSL 3.5.0 this is not longer required and causes compilation failures.


orbea:
@vgough I realize this project is not maintained, but I would really appreciate if you could review this relatively trivial PR.

vgough:
sorry, the C++ build is no more.

PR #705, opened 2026-08-10
build and test on FreeBSD

neilpang:
The Cirrus FreeBSD task last ran on 2026-04-01. Before that it had
failed on every commit since the Rust port landed -- 28 runs, none green
-- and allow_failures: true reported each one as neutral, so nothing
surfaced. .cirrus.yml was last edited on 2026-07-25, after it had
already stopped running.

This adds .github/workflows/freebsd.yml, shaped like ci.yml: the same
clippy, build, test and live-mount sequence. The VM boots under QEMU on
an ordinary ubuntu-latest runner, so there is no service, no dedicated
runner and no compute credits.

Package names come from the pkg database on the box rather than from
memory: fusefs-libs3 supplies fuse3.pc 3.18.1, protobuf supplies protoc,
llvm supplies the libclang bindgen needs, and fusefs.ko is already in
GENERIC so nothing has to kldload it. .cirrus.yml installs fusefs-libs,
which is FUSE 2 and carries no fuse3.pc, so that task could not have
built even while it was still running.

build.rs

protoc-bin-vendored ships binaries for Linux, macOS and Windows only, so
protoc_bin_path() returns Err on FreeBSD and .expect() panics. Leaving
PROTOC unset there lets prost-build find protoc on PATH instead.

src/reverse_fs.rs

Four uses of libc::ENODATA, which FreeBSD does not define. Every one
means "extended attribute not found", so they become Errno::ENOATTR,
which typed-fuse already gates: on Linux that IS libc::ENODATA, so the
change is bit-identical there.

the xattr name encoding -- the one worth reading

Stored attribute names are base64 of the encrypted name, and the
standard alphabet contains '/', which FreeBSD will not accept in an
attribute name. Four names differing in a single character:

    setextattr user "encfs.abcdef"  v f    rc=0
    setextattr user "encfs.abc+def" v f    rc=0
    setextattr user "encfs.abc=def" v f    rc=0
    setextattr user "encfs.abc/def" v f    rc=1  Invalid argument

Of the six distinct names this repo's own xattr tests generate, exactly
the two containing '/' are the two whose setxattr returned EINVAL:

    user.encfs.LxjRlnYLUZf/KnISEoYMnw    user.empty
    user.encfs.1C/+DS/JPpFmEBmUPPJXgw    user.simple

Ten of the names contained '+' and all stored fine. So on FreeBSD
roughly a third of extended attributes cannot be stored at all.

New names use the URL-safe alphabet, which spells the two disputed
characters '-' and '_'. Reads accept either, and getxattr and
removexattr fall back to the old spelling, so attributes written by
earlier builds stay reachable. The alphabets differ only in those four
characters, so a string that decodes under both contains none of them
and yields the same bytes either way: trying one and then the other
cannot return the wrong plaintext.

Two things this does not touch. Filenames go through
SslCipher::filename_base64_encode, which uses encfs's own alphabet, not
this one. And there is no compatibility to keep with the C++ encfs: it
passed attribute names to the backing file unchanged, so the encrypted
and encoded scheme has only ever existed in this port.

tests

tests/live/mod.rs defined mountinfo_has_mount for Linux and macOS only.
FreeBSD has no /proc either and its mount(8) prints the same
"<device> on <mountpoint> (<fstype>, ...)" shape, so it joins the macOS
arm.

tests/xattr_test.rs read the backing names with llistxattr; FreeBSD gets
an extattr_list_link arm. There the namespace is not part of the name,
so what is stored is "encfs.<b64>", and the assertion says so instead of
being made to look uniform.

tests/permissions_test.rs asserted symlink mode 0o777, already excluding
macOS. Linux fixes those bits and ignores them; the BSDs give symlinks
real mode bits -- FreeBSD ships lchmod(2), "similar to chmod() but does
not follow symbolic links" -- and report 0o755, so FreeBSD joins the
exclusion.

Verified

Linux, this branch as it stands, under ci.yml: 215 tests across every
binary, 0 failed, including 29 live mount tests. No companion change
needed for this half.
https://github.com/neilpang/encfs/actions/runs/31366219652

FreeBSD 15.1: build, clippy -D warnings, 5/5 xattr, 4/4 permissions,
29/29 live mount. That run used these source files byte for byte plus a
temporary Cargo.toml patch pointing typed-fuse at the companion branch;
the patch is not in this PR.
https://github.com/neilpang/encfs/actions/runs/31364870397

Sequencing

The FreeBSD job goes green once vgough/typed-fuse#3 is in. Without it
the job stops here:

  Compiling typed-fuse v0.6.0 (https://github.com/vgough/typed-fuse#dd70e273)
  error[E0425]: cannot find value `ENODATA` in crate `libc`

https://github.com/neilpang/encfs/actions/runs/31366219628

typed-fuse does not build on FreeBSD today: no l*xattr syscalls, an
ungated ENODATA, a missing struct flock field, and a fuse_session_new
that libfuse 3.17 turned into a macro. Cargo.toml tracks its default
branch with no rev pin, so this picks the fix up on its own.

Two loose ends, yours to call: .cirrus.yml is dead weight now, and
ci.yml pins actions/checkout@v4 while v7 is current.

Written with the help of Claude; I read and tested all of it.


.github/workflows/freebsd.yml
```diff
@@ -0,0 +1,50 @@
...
+      uses: vmactions/freebsd-vm@v1
+      timeout-minutes: 90
+      with:
+        release: '15'
+        usesh: true
+        copyback: false
+        cache-after-prepare: true
+
+        prepare: |
+          pkg install -y git pkgconf rust protobuf llvm fusefs-libs3
```
vgough, line 31: Does this also need a `kldload fusefs` ?

vgough: added it, CI is passing.

vgough (approved):
Thanks!

